PR #9, opened 2023-09-21
test(output): compare python output to rust output

LucasLvy:


tests/common.rs
```diff
@@ -37,3 +37,28 @@ pub fn setup_pie() -> CairoPie {
 
     runner.get_cairo_pie(&vm).unwrap()
 }
+
+#[allow(unused)]
```
EvolveArt, line 41: why is it flag as `unused` ?

tests/common.rs
```diff
@@ -37,3 +37,28 @@ pub fn setup_pie() -> CairoPie {
...
+    let python_output = python_output
+        .split('\n')
+        .into_iter()
+        .skip_while(|&x| x != "Program output:")
+        .skip(1)
+        .filter(|&x| !x.trim().is_empty())
+        .into_iter();
+    for (rs, py) in rs_output.zip(python_output) {
+        let py = py.to_string().trim().to_string();
+        assert_eq!(*rs, py);
```
EvolveArt, line 62: let's use pretty_assert

tests/snos.rs
```diff
@@ -54,6 +57,7 @@ fn custom_hint_ok() {
         &mut hint_processor,
     )
     .expect("Couldn't run program");
+    compare_python_output(program_path, virtual_machine);
 }
```
EvolveArt, line 61: let's add a test for a case where it fails too, we need better error handling or it's gonna be a mess later on

EvolveArt (changes requested):
left few comments

tests/common.rs
LucasLvy: no clue but it's very annoying. But i don't want to have it in the same file as the tests

.github/workflows/ci.yml
```diff
@@ -28,4 +28,4 @@ jobs:
         run: rustup show
       - uses: Swatinem/rust-cache@v2
       - name: Run tests
-        run: cargo test
+        run: pip install cairo-lang; cargo test
```
EvolveArt, line 31: Cam we setup python cleanly in the repo and just use the submodule cairo-lang instead of the pip one ?

tests/common.rs
```diff
@@ -37,3 +37,28 @@ pub fn setup_pie() -> CairoPie {
...
+    let python_output = python_output
+        .split('\n')
+        .into_iter()
+        .skip_while(|&x| x != "Program output:")
+        .skip(1)
+        .filter(|&x| !x.trim().is_empty())
+        .into_iter();
+    for (i, (rs, py)) in rs_output.zip(python_output).enumerate() {
+        let py = py.to_string().trim().to_string();
+        pretty_assertions::assert_eq!(*rs, py, "Output #{i:} is different");
```
EvolveArt, line 62: please create an issue to improve that

tests/common.rs
```diff
@@ -37,3 +37,28 @@ pub fn setup_pie() -> CairoPie {
 
     runner.get_cairo_pie(&vm).unwrap()
 }
+
+#[allow(unused)]
+pub fn compare_python_output(program: &str, mut vm: VirtualMachine) {
```
EvolveArt, line 42:
```suggestion
pub fn assert_python_and_rust_output_match(program: &str, mut vm: VirtualMachine) {
```

EvolveArt (changes requested):

.github/workflows/ci.yml
LucasLvy: it's gonna be way slower so i'd rather not

tests/common.rs
LucasLvy: nothing needs to be improved it's perfect

.github/workflows/ci.yml
EvolveArt: wdym? if we just checkout using the submodule no need to install anything even no?

EvolveArt (approved):

.github/workflows/ci.yml
LucasLvy: It’s not that easy check the docker file in cairo lang to see what are the steps to build the package from source

PR #22, opened 2023-09-26
feat(hints): add load compiled classes

0xLucqs:


tests/snos.rs
```diff
@@ -30,6 +34,22 @@ fn print_a_hint(
     Ok(())
 }
 
+/// Mocks an os input but not a good way of doing things as it won't work in python.
+fn mock_os_input(
```
EvolveArt, line 38: let's move it to a `mock` module where we will add the scope mocking functionnality

tests/snos.rs
```diff
@@ -98,3 +118,35 @@ fn test_different_outputs() {
     .expect("Couldn't run program");
     assert_python_and_rust_output_match("build/different_output.json", virtual_machine);
 }
+
+#[test]
+fn load_compiled_classes_facts_test() {
```
EvolveArt, line 123: could we externalize the logic of this function by extending the SnOsRunner ? I feel like we're repeating ourselves here and makes it hard to debug

drspacemn (commented):
warning: unused import: `cairo_vm::hint_processor::builtin_hint_processor::hint_utils::get_integer_from_var_name`
 --> tests/snos.rs:7:5
  |
7 | use cairo_vm::hint_processor::builtin_hint_processor::hint_utils::get_integer_from_var_name;
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: `#[warn(unused_imports)]` on by default

tests/contracts/hint.cairo
```diff
@@ -4,8 +4,6 @@ func main(output_ptr: felt*) -> (output_ptr: felt*) {
     tempvar a = 17;
     a = [output_ptr], ap++;
 
-    // Use custom hint to print the value of a
-    %{ print(ids.a) %}
```
drspacemn, line 8: isn't this the purpose of the `hint.cairo` file?

tests/snos.rs
drspacemn: agreed. To handle in follow up.

PR #23, opened 2023-09-27
dev: build script, remove async for now, remove build artifacts

drspacemn:
- build script
- remove async for now 
- remove build artifacts

once we identify a place that we can utilize threading we can add back in async


tests/snos.rs
```diff
@@ -81,20 +77,15 @@ fn test_different_outputs() {
...
+        program_content,
         &CairoRunConfig {
             layout: "all_cairo",
             ..Default::default()
         },
         &mut hint_processor,
     )
-    .expect("Couldn't run program");
-    assert_python_and_rust_output_match("build/different_output.json", virtual_machine);
+    .expect("Couldn't run programm");
```
EvolveArt, line 89: typo

EvolveArt (commented):
Shouldn't we recompile even if the compiled files already exist ? At least make sure they haven't changed

drspacemn:
> Shouldn't we recompile even if the compiled files already exist ? At least make sure they haven't changed

In this case the GH Action doesn't cache the build dir so it will recompile the test files each time. IMO, it's a better dev experience if we don't recompile a contract that is already in the build dir.

If you've made a change to that contract simply delete the compiled contract from build and re run the tests.

EvolveArt:
> > Shouldn't we recompile even if the compiled files already exist ? At least make sure they haven't changed
> 
> In this case the GH Action doesn't cache the build dir so it will recompile the test files each time. IMO, it's a better dev experience if we don't recompile a contract that is already in the build dir locally.
> 
> If you've made a change to that contract simply delete the compiled contract from build and re-run the tests.

Ok will just need to be careful if we add cache

PR #42, opened 2023-10-23
Lucas/load tx

LucasLvy:


drspacemn (approved):
moved the os output stuff to a new branch, so pragma can work w/ the tx input stuff.

PR #44, opened 2023-10-25
More hints

apoorvsadana:


src/hints/mod.rs
```diff
@@ -259,9 +277,87 @@ pub fn load_next_tx(
...
+    let constructor_calldata_size = match &tx.constructor_calldata {
+        None => 0,
+        Some(calldata) => calldata.len(),
+    };
+    insert_value_from_var_name("constructor_calldata_size", constructor_calldata_size, vm, ids_data, ap_tracking)?;
+
+    let constructor_calldata: Vec<MaybeRelocatable> = match tx.constructor_calldata {
+        None => vec![],
+        Some(calldata) => calldata.iter().map(|felt| felt.into()).collect(),
+    };
```
LucasLvy, line 321: Is that needed? Can't you just do `calldata.iter().map(|felt| felt.into()).collect()` and it'd be an empty vec if the length is 0?

src/hints/mod.rs
```diff
@@ -259,9 +277,87 @@ pub fn load_next_tx(
...
+/// ids.constructor_calldata = segments.gen_arg(arg=tx.constructor_calldata)
+pub fn prepare_constructor_execution(
+    vm: &mut VirtualMachine,
+    exec_scopes: &mut ExecutionScopes,
+    ids_data: &HashMap<String, HintReference>,
+    ap_tracking: &ApTracking,
+    _constants: &HashMap<String, Felt252>,
+) -> Result<(), HintError> {
+    let tx = exec_scopes.get::<InternalTransaction>("tx")?;
+    // Safe to unwrap because the remaining number of txs is checked in the cairo code.
```
LucasLvy, line 302: Doesn't sound like the right comment ?

src/hints/mod.rs
```diff
@@ -259,9 +277,87 @@ pub fn load_next_tx(
...
+    insert_value_from_var_name(
+        "contract_address_salt",
+        tx.contract_address_salt.unwrap_or(Felt252::from(0)),
```
LucasLvy, line 305: if it's safe, no need for the `unwrap_or` and probably we want it to panic so it's easier to debug instead of a 0

src/hints/mod.rs
```diff
@@ -259,9 +277,87 @@ pub fn load_next_tx(
...
+        vm,
+        ids_data,
+        ap_tracking,
+    )?;
+    insert_value_from_var_name("class_hash", tx.class_hash.unwrap_or(Felt252::from(0)), vm, ids_data, ap_tracking)?;
```
LucasLvy, line 310: same here

LucasLvy (changes requested):

src/hints/mod.rs
apoorvsadana: I am not sure what you meant here, if you were talking about removing `match` part then I can't do that because it's a `Option`. So I can't `iter` directly. However, I did modify the code to make it smaller and use `unwrap_or_default`

src/hints/mod.rs
apoorvsadana: Yep, my bad, fixed

src/hints/mod.rs
apoorvsadana: This one seems to be safe for the existing input so I have added an `expect`. Is there more information about what `input.json` is about? I am not very sure what the input is trying to do right now

src/hints/mod.rs
apoorvsadana: This one fails if I don't return 0 - so class_hash is needed but is not available. The above information would help here as well

src/hints/mod.rs
LucasLvy: the `input.json` is the input given to the os

src/hints/mod.rs
drspacemn: this still need to be changed to `contract_hash` to match the input JSON

drspacemn (approved):
lgtm

PR #48, opened 2023-11-02
format_output_cairo debugging

apoorvsadana:


drspacemn:
@maciejka will be looking into a more refined way to run unit tests.

Those will be cover this piece of code

PR #50, opened 2023-12-06
Poetry config, typos

maciejka:


maciejka:
Any ideas why tests fail? @drspacemn

LucasLvy:
poetry might be a bit too much here no? Can't we just have a `requirements.txt`?

maciejka:
> poetry might be a bit too much here no? Can't we just have a `requirements.txt`?

The only advantage I see is that poetry manages python version explicitly.

PR #52, opened 2023-12-12
List hints again

maciejka:
Choose which hints subset you are interested in:
```
cargo run --bin hints all | implemented | unimplemented
```


LucasLvy:
you need to consider the base hints that are already implemented in the vm (you can use the starknet whitelisted hints file to get those)

LucasLvy:
and i'd display the number of hints not the actual code

maciejka:
> and i'd display the number of hints not the actual code

You can use `jq` to process it the way you want:
```
cargo run --bin hints unimplemented | jq length
```

drspacemn:
Having trouble understanding the purpose of this? This is to display the meta-data for the current hint implementations in SNOS? 

Or simply to output their python strings in a large array?

Since this is meant to be used as a library I'm hesitant to add a top level `bin/` dir that doesn't add to the function of snos

drspacemn (commented):
Having trouble understanding the purpose of this? This is to display the meta-data for the current hint implementations in SNOS?

Or simply to output their python strings in a large array?

Since this is meant to be used as a library I'm hesitant to add a top level bin/ dir that doesn't add to the function of snos

For hint accounting or metadata on dev progress I'd probably like to keep that to `scripts || tests || benches`

maciejka:
It is a utility like a script. Ultimately it can be used in ci script to detect unimplemented or no longer needed hints.
The scripts dir seems most natural to me. Ok?

drspacemn:
> The scripts dir seems most natural to me. Ok?

yes that sounds perfect to me

src/hints/mod.rs
```diff
@@ -33,96 +36,53 @@ use crate::io::InternalTransaction;
...
-
-    let is_deprecated_hint = HintFunc(Box::new(is_deprecated));
-    hint_processor.add_hint(String::from(IS_DEPRECATED), Rc::new(is_deprecated_hint));
-
-    let os_context_segments_hint = HintFunc(Box::new(os_context_segments));
-    hint_processor.add_hint(String::from(OS_CONTEXT_SEGMENTS), Rc::new(os_context_segments_hint));
-
-    let selected_builtins_hint = HintFunc(Box::new(selected_builtins));
-    hint_processor.add_hint(String::from(SELECTED_BUILTINS), Rc::new(selected_builtins_hint));
+pub fn sn_hint_processor() -> BuiltinHintProcessor {
```
drspacemn, line 47: #54

Will re-init how we set this handler up, so maybe we should hold off on this portion until that is merged

src/hints/mod.rs
```diff
@@ -24,16 +24,65 @@ use starknet_api::deprecated_contract_class::ContractClass as DeprecatedContract
...
+    &mut ExecutionScopes,
+    &HashMap<String, HintReference>,
+    &ApTracking,
+) -> Result<HintExtension, HintError>;
+
+static EXTENSIVE_HINTS: [(&str, ExtensiveHintImpl); 1] = [(CHECK_DEPRECATED_CLASS_HASH, check_deprecated_class_hash)];
+
+pub struct SnosHintProcessor<'a> {
+    builtin_hint_proc: BuiltinHintProcessor,
+    hints: HashMap<&'a str, HintImpl>,
```
drspacemn, line 80: Let's use `String` and get rid of the lifetime

drspacemn (approved):

drspacemn (changes requested):
Let's use `String` to get rid of the lifetime

drspacemn (approved):

PR #84, opened 2024-02-26
Run os test on a simple block

maciejka:
Issue Number: N/A

## Type

- [ ] feature
- [ ] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [ ] yes
- [ ] no


src/hints/block_context.rs
```diff
@@ -137,15 +108,15 @@ pub fn load_deprecated_class(
...
-    // ap_tracking)?; let computed_hash = vm.get_integer(computed_hash_addr)?;
-    // let expected_hash = exec_scopes.get::<Felt252>("compiled_class_hash").unwrap();
-    // if computed_hash.as_ref() != &expected_hash {
-    //     return Err(HintError::AssertionFailed(
-    //         format!("Compiled_class_hash mismatch comp={computed_hash}
-    // exp={expected_hash}").into_boxed_str(),     ));
-    // }
+    let computed_hash_addr = get_ptr_from_var_name("compiled_class_fact", vm, ids_data, ap_tracking)?;
+    let computed_hash = vm.get_integer(computed_hash_addr)?;
+    let expected_hash = exec_scopes.get::<Felt252>("compiled_class_hash").unwrap();
```
odesenfans, line 113: For variable names, we introduced the `vars` module and its `ids` and `scopes` submodules. Could you define constants there instead of using hardcoded strings?

src/hints/block_context.rs
```diff
@@ -137,15 +108,15 @@ pub fn load_deprecated_class(
...
+
+    if computed_hash.as_ref() != &expected_hash {
+        return Err(HintError::AssertionFailed(
+            format!("Compiled_class_hash mismatch comp={computed_hash} exp={expected_hash}").into_boxed_str(),
```
odesenfans, line 117: Could you use the same error message as the original hint? I think it's better to stay as close as possible to the original.

src/hints/block_context.rs
```diff
@@ -176,6 +147,141 @@ pub fn load_deprecated_class(
...
+    ap_tracking: &ApTracking,
+    _constants: &HashMap<String, Felt252>,
+) -> Result<(), HintError> {
+    let os_input = exec_scopes.get::<StarknetOsInput>("os_input")?;
+    let compiled_class_facts_ptr = vm.add_memory_segment();
+    insert_value_from_var_name("compiled_class_facts", compiled_class_facts_ptr, vm, ids_data, ap_tracking)?;
+
+    insert_value_from_var_name("n_compiled_class_facts", os_input.compiled_classes.len(), vm, ids_data, ap_tracking)?;
+
+    let compiled_class_facts: Box<dyn Any> = Box::new(os_input.compiled_classes.into_iter());
```
odesenfans, line 171: Doesn't that create a `Box<Iterator>`? I'd collect the result into a vector just to protect us in case of multiple iterations. 

Edit: Saw the code below, seems like that's the behaviour intended in the Python VM as well. Correct?

src/hints/block_context.rs
```diff
@@ -176,6 +147,141 @@ pub fn load_deprecated_class(
...
+pub fn load_class_inner(
+    vm: &mut VirtualMachine,
+    exec_scopes: &mut ExecutionScopes,
+    ids_data: &HashMap<String, HintReference>,
+    ap_tracking: &ApTracking,
+    _constants: &HashMap<String, Felt252>,
+) -> Result<(), HintError> {
+    let class_iter = exec_scopes.get_mut_ref::<IntoIter<Felt252, CasmContractClass>>("compiled_class_facts")?;
+
+    let (compiled_class_hash, class) = class_iter.next().unwrap();
```
odesenfans, line 211: Isn't this `unwrap()`  bound to panic at the end of the iterator?

src/hints/block_context.rs
```diff
@@ -176,6 +147,141 @@ pub fn load_deprecated_class(
...
+    ids_data: &HashMap<String, HintReference>,
+    ap_tracking: &ApTracking,
+) -> Result<HintExtension, HintError> {
+    let compiled_class_fact_addr = get_relocatable_from_var_name("compiled_class_fact", vm, ids_data, ap_tracking)?;
+    let computed_hash = vm.get_integer(compiled_class_fact_addr)?; //TODO: calculate .hash explicitly
+    let expected_hash = exec_scopes.get::<Felt252>("compiled_class_hash").unwrap();
+
+    if computed_hash.as_ref() != &expected_hash {
+        return Err(HintError::AssertionFailed(
+            format!("Compiled_class_hash mismatch comp={computed_hash} exp={expected_hash}").into_boxed_str(),
```
odesenfans, line 266: Same, keep the error messages similar to the Python ones.

src/hints/block_context.rs
```diff
@@ -176,6 +147,141 @@ pub fn load_deprecated_class(
...
+};
+pub fn load_class(
+    _hint_processor: &dyn HintProcessor,
+    vm: &mut VirtualMachine,
+    exec_scopes: &mut ExecutionScopes,
+    ids_data: &HashMap<String, HintReference>,
+    ap_tracking: &ApTracking,
+) -> Result<HintExtension, HintError> {
+    let compiled_class_fact_addr = get_relocatable_from_var_name("compiled_class_fact", vm, ids_data, ap_tracking)?;
+    let computed_hash = vm.get_integer(compiled_class_fact_addr)?; //TODO: calculate .hash explicitly
```
odesenfans, line 261: Nitpick: Shouldn't you define `compiled_class_fact` as a struct that derives `CairoType`?

src/hints/block_context.rs
```diff
@@ -286,3 +392,47 @@ pub fn get_block_mapping(
...
+    vm: &mut VirtualMachine,
+    _exec_scopes: &mut ExecutionScopes,
+    ids_data: &HashMap<String, HintReference>,
+    ap_tracking: &ApTracking,
+    _constants: &HashMap<String, Felt252>,
+) -> Result<(), HintError> {
+    // let bytecode_segment_structure = get_ptr_from_var_name("bytecode_segment_structure", vm,
+    // ids_data, ap_tracking)?; TODO: complete when bytecode_segment_structure is implemented
+    insert_value_from_var_name("is_leaf", 1, vm, ids_data, ap_tracking)
+}
```
odesenfans, line 438: I think there should be unit tests for each hint that is already fully implemented. Implementing tests for all the hints at the end is gonna be such a pain otherwise.

src/hints/execution.rs
```diff
@@ -326,3 +361,409 @@ pub fn exit_call(
...
+
+pub fn contract_address(
+    vm: &mut VirtualMachine,
+    exec_scopes: &mut ExecutionScopes,
+    ids_data: &HashMap<String, HintReference>,
+    ap_tracking: &ApTracking,
+    _constants: &HashMap<String, Felt252>,
+) -> Result<(), HintError> {
+    let tx = exec_scopes.get::<InternalTransaction>("tx")?;
+    let contract_address = if tx.r#type == "L1_HANDLER" {
```
odesenfans, line 382: Out of curiosity, what is this `tx.r#type` syntax?

odesenfans (changes requested):
A first pass, will review the rest after lunch.

src/hints/block_context.rs
maciejka: Right. It is not any more fragile than the rest of this cairo/imperative language construction.

src/hints/block_context.rs
maciejka: End of the iterator won't be reached it everything goes well (iteration is driven from cairo). Added more explicit error message.

src/hints/block_context.rs
maciejka: I would say that reasonable middle ground is this: If hints are straightforward there is no need for unit tests since this code will be executed by integration tests. I would write unit tests only for non trivial hints involving ifs/loops/complex data structures.

src/hints/execution.rs
maciejka: Not sure. `r#type` is an identifier. Not written by me, maybe just a typo?

src/hints/execution.rs
```diff
@@ -326,3 +361,409 @@ pub fn exit_call(
...
+    //     vm.load_data(resource_bounds_base, &resource_bounds)?;
+    //     resource_bounds_base
+    // };
+
+    let resource_bounds = 0;
+    insert_value_from_var_name("resource_bounds", resource_bounds, vm, ids_data, ap_tracking)
+}
+
+// copilot prompt:
+// implement rust hints for the following python hints:
```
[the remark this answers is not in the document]
maciejka: You mean we will forget about it?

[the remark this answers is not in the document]
notlesh: Yeah, I imagine it getting forgotten about, getting stale, and getting out of sync with the hints it is about, and I think an issue about it could be a good way to make sure we revisit it later.

tests/common/block_utils.rs
```diff
@@ -0,0 +1,247 @@
...
+        FeatureContract::ERC20 => {
+            let fe = FieldElement::from_dec_str(
+                "2553874082637258309275750418379019378586603706497644242041372159420778949015",
+            )
+            .unwrap();
+            StarkHash::from(fe)
+        }
+        // FeatureContract::FaultyAccount(_) => FAULTY_ACCOUNT_BASE,
+        // FeatureContract::LegacyTestContract => LEGACY_CONTRACT_BASE,
+        // FeatureContract::SecurityTests => SECURITY_TEST_CONTRACT_BASE,
```
odesenfans, line 87: Why are all these lines commented?

tests/common/blocks.rs
```diff
@@ -0,0 +1,111 @@
...
+            contract_address,
+            "return_result",
+            &[stark_felt!(2_u8)],
+        ),
+        version: tx_version,
+        nonce: nonce_manager.next(account_address),
+        only_query,
+    });
+
+    let account_tx_intenal = to_internal_tx(&account_tx);
```
odesenfans, line 62: Typo: intenal -> internal

odesenfans (changes requested):
Some more comments (minor). For the rest there  are too many TODOs, commented code etc. I'll trust you on mainlining this and we can improve things with smaller PRs afterwards.

tests/common/block_utils.rs
maciejka: Whole test setup is work in progress so I treat it as a draft leaving code commented out in case it might be useful later.

src/hints/execution.rs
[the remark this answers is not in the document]
maciejka: We won't forget it because we know we need to support transactions v3 and we have a TODOs that are easy to track:
image
.

PR #87, opened 2024-03-12
Feature: port of all Patricia trie related functions from cairo-lang

odesenfans:
This port brings over the PatriciaTree, BinaryTreeNodeFact,
Calculation, LeafFact, etc logic from cairo-lang.

In short, this PR allows to compute state tree roots, state tree
updates, state tree diffs, etc while staying as close as possible to the
Python implementation.

Known limitations/differences:
* Tree traversal (`traverse_tree`) is single threaded. The Python VM
  spawns worker tasks for traversal but this approach proved to be too
  complex to implement in Rust for a first go.
* Error handling is not perfectly defined, lots of functions resort to
  TreeError but the definition of errors inside this enum could be
  improved.

## Type

- [x] feature
- [ ] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [ ] yes
- [x] no

### DIFF
--- a/src/starkware_utils/commitment_tree/binary_fact_tree.rs
+++ b/src/starkware_utils/commitment_tree/binary_fact_tree.rs
@@ -12,14 +12,14 @@ pub trait Leaf: Clone {}
 pub type BinaryFactDict = HashMap<BigUint, Vec<BigUint>>;
 
 #[allow(async_fn_in_trait)]
-pub trait BinaryFactTree<S, H, L>: Sized
+pub trait BinaryFactTree<S, H, LF>: Sized
 where
     S: Storage,
     H: HashFunctionType,
-    L: LeafFact<S, H>,
+    LF: LeafFact<S, H>,
 {
     /// Initializes an empty BinaryFactTree of the given height.
-    async fn empty_tree(ffc: &mut FactFetchingContext<S, H>, height: Height, leaf_fact: L) -> Result<Self, TreeError>;
+    async fn empty_tree(ffc: &mut FactFetchingContext<S, H>, height: Height, leaf_fact: LF) -> Result<Self, TreeError>;
 
     /// Returns the values of the leaves whose indices are given.
     async fn get_leaves(
@@ -27,9 +27,9 @@ where
         ffc: &mut FactFetchingContext<S, H>,
         indices: &[TreeIndex],
         facts: &mut Option<BinaryFactDict>,
-    ) -> Result<HashMap<TreeIndex, L>, TreeError>;
+    ) -> Result<HashMap<TreeIndex, LF>, TreeError>;
 
-    async fn get_leaf(&self, ffc: &mut FactFetchingContext<S, H>, index: TreeIndex) -> Result<Option<L>, TreeError> {
+    async fn get_leaf(&self, ffc: &mut FactFetchingContext<S, H>, index: TreeIndex) -> Result<Option<LF>, TreeError> {
         let mut facts = None;
         let leaves = self.get_leaves(ffc, vec![index.clone()].as_ref(), &mut facts).await?;
         Ok(leaves.get(&index).cloned())
@@ -43,7 +43,7 @@ where
     async fn update(
         &mut self,
         ffc: &mut FactFetchingContext<S, H>,
-        modifications: Vec<(TreeIndex, L)>,
+        modifications: Vec<(TreeIndex, LF)>,
         facts: &mut Option<BinaryFactDict>,
     ) -> Result<Self, TreeError>;
 }
--- a/src/starkware_utils/commitment_tree/binary_fact_tree.rs
+++ b/src/starkware_utils/commitment_tree/binary_fact_tree.rs
@@ -1,0 +1,49 @@
+use std::collections::HashMap;
+
+use num_bigint::BigUint;
+
+use crate::starkware_utils::commitment_tree::base_types::{Height, TreeIndex};
+use crate::starkware_utils::commitment_tree::errors::TreeError;
+use crate::starkware_utils::commitment_tree::leaf_fact::LeafFact;
+use crate::storage::storage::{FactFetchingContext, HashFunctionType, Storage};
+
+pub trait Leaf: Clone {}
+
+pub type BinaryFactDict = HashMap<BigUint, Vec<BigUint>>;
+
+#[allow(async_fn_in_trait)]
+pub trait BinaryFactTree<S, H, L>: Sized
+where
+    S: Storage,
+    H: HashFunctionType,
+    L: LeafFact<S, H>,
+{
+    /// Initializes an empty BinaryFactTree of the given height.
+    async fn empty_tree(ffc: &mut FactFetchingContext<S, H>, height: Height, leaf_fact: L) -> Result<Self, TreeError>;
+
+    /// Returns the values of the leaves whose indices are given.
+    async fn get_leaves(
+        &self,
+        ffc: &mut FactFetchingContext<S, H>,
+        indices: &[TreeIndex],
+        facts: &mut Option<BinaryFactDict>,
+    ) -> Result<HashMap<TreeIndex, L>, TreeError>;
+
+    async fn get_leaf(&self, ffc: &mut FactFetchingContext<S, H>, index: TreeIndex) -> Result<Option<L>, TreeError> {
+        let mut facts = None;
+        let leaves = self.get_leaves(ffc, vec![index.clone()].as_ref(), &mut facts).await?;
+        Ok(leaves.get(&index).cloned())
+    }
+
+    /// Updates the tree with the given list of modifications, writes all the new facts to the
+    /// storage and returns a new BinaryFactTree representing the fact of the root of the new tree.
+    ///
+    /// If facts argument is not None, this dictionary is filled during traversal through the tree
+    /// by the facts of their paths from the leaves up.
+    async fn update(
+        &mut self,
+        ffc: &mut FactFetchingContext<S, H>,
+        modifications: Vec<(TreeIndex, L)>,
+        facts: &mut Option<BinaryFactDict>,
+    ) -> Result<Self, TreeError>;
+}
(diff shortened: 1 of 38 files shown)


src/starkware_utils/commitment_tree/binary_fact_tree.rs
```diff
@@ -0,0 +1,49 @@
...
+use crate::starkware_utils::commitment_tree::errors::TreeError;
+use crate::starkware_utils::commitment_tree::leaf_fact::LeafFact;
+use crate::storage::storage::{FactFetchingContext, HashFunctionType, Storage};
+
+pub trait Leaf: Clone {}
+
+pub type BinaryFactDict = HashMap<BigUint, Vec<BigUint>>;
+
+#[allow(async_fn_in_trait)]
+pub trait BinaryFactTree<S, H, LF>: Sized
```
notlesh, line 15: A brief comment about this trait would help

src/starkware_utils/commitment_tree/binary_fact_tree_node.rs
```diff
@@ -0,0 +1,322 @@
...
+use num_bigint::BigUint;
+
+use crate::starkware_utils::commitment_tree::base_types::{Height, TreeIndex};
+use crate::starkware_utils::commitment_tree::binary_fact_tree::BinaryFactDict;
+use crate::starkware_utils::commitment_tree::errors::TreeError;
+use crate::starkware_utils::commitment_tree::inner_node_fact::InnerNodeFact;
+use crate::starkware_utils::commitment_tree::leaf_fact::LeafFact;
+use crate::starkware_utils::commitment_tree::merkle_tree::traverse_tree::{traverse_tree, TreeTraverser};
+use crate::storage::storage::{FactFetchingContext, HashFunctionType, Storage, StorageError};
+
```
notlesh, line 14: Some comments or pointers in general in this file would help, or perhaps keeping things `pub(crate)` if they aren't meant to be used externally...

src/starkware_utils/commitment_tree/binary_fact_tree_node.rs
```diff
@@ -0,0 +1,322 @@
...
+    /// current BinaryFactTreeNode.
+    ///
+    /// If facts argument is not None, this dictionary is filled with facts read from the DB.
+    fn get_children<'a>(
+        &'a self,
+        ffc: &'a mut FactFetchingContext<S, H>,
+        facts: &'a mut Option<BinaryFactDict>,
+    ) -> impl std::future::Future<Output = Result<(Self, Self), TreeError>> + Send;
+
+    fn _get_leaves<'a>(
```
notlesh, line 169: is the `_foo()` convention here simply to keep things as close to the Python implementation as possible? Seems this is implied by not being `pub` in rust, but I also see value in keeping things identical for now

notlesh: Oh, this is a `trait`... :thinking:

src/starkware_utils/commitment_tree/merkle_tree/traverse_tree.rs
```diff
@@ -0,0 +1,48 @@
+use std::collections::VecDeque;
+
+use crate::starkware_utils::commitment_tree::errors::TreeError;
+use crate::starkware_utils::commitment_tree::leaf_fact::LeafFact;
+use crate::storage::storage::{HashFunctionType, Storage};
+
+pub trait TreeTraverser<'test, S, H, LF>
```
notlesh, line 7: why `'test`?

src/starkware_utils/commitment_tree/patricia_tree/patricia_utils.rs
```diff
@@ -0,0 +1,173 @@
...
+pub struct Node {
+    bottom: Felt252,
+    path: Felt252,
+    length: Felt252,
+}
+
+impl Node {
+    fn empty() -> Self {
+        Self { bottom: Felt252::ZERO, path: Felt252::ZERO, length: Felt252::ZERO }
+    }
```
notlesh, line 36: Could be better expressed as a `Default` impl

src/starkware_utils/commitment_tree/patricia_tree/patricia_utils.rs
```diff
@@ -0,0 +1,173 @@
...
+
+    fn is_empty(&self) -> bool {
+        let empty_node = Self::empty();
+        self.bottom == empty_node.bottom && self.path == empty_node.path && self.length == empty_node.length
```
notlesh, line 40: Could impl `PartialEq` instead

src/starkware_utils/commitment_tree/patricia_tree/virtual_calculation_node.rs
```diff
@@ -0,0 +1,650 @@
...
+pub struct BinaryCalculation<S, H, LF>
+where
+    S: Storage + 'static,
+    H: HashFunctionType + 'static,
+    LF: LeafFact<S, H>,
+{
+    left: Box<HashCalculationImpl<S, H, LF>>,
+    right: Box<HashCalculationImpl<S, H, LF>>,
+    _s: PhantomData<S>,
+    _h: PhantomData<H>,
```
notlesh, line 32: You can use a tuple: `_phantom: PhantomData<(S, H)>`

src/starkware_utils/commitment_tree/calculation.rs
```diff
@@ -0,0 +1,276 @@
...
+    fn default() -> Self {
+        Self { inner_nodes: Default::default(), leaves: Default::default() }
+    }
+}
+
+/// A calculation that can produce a result of type T. The calculation is dependent on the results
+/// of other calculations. Those calculations can be of type other than T.
+/// The result of the calculation can be produced when the results of the dependency calculations
+/// are given.
+pub trait Calculation<T, LF> {
```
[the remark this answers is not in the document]
notlesh: Seems like this is used primarily for hashing, it might help to call that example out in the documentation (or rename the trait if that makes sense)

src/starkware_utils/commitment_tree/patricia_tree/virtual_patricia_node.rs
```diff
@@ -0,0 +1,690 @@
...
+    pub bottom_node: Vec<u8>,
+    pub path: NodePath,
+    pub length: Length,
+    /// The height of the subtree rooted at this node.
+    /// In other words, this is the length of the path from this node to the leaves.
+    pub height: Height,
+
+    _s: PhantomData<S>,
+    _h: PhantomData<H>,
+    _l: PhantomData<LF>,
```
notlesh, line 36: Same as above, could use a tuple

src/starkware_utils/commitment_tree/update_tree.rs
```diff
@@ -0,0 +1,375 @@
...
+    S: Storage + Sync + Send,
+    H: HashFunctionType + Sync + Send,
+    LF: LeafFact<S, H>,
+    CN: CalculationNode<S, H, LF>,
+{
+    let mut cur_node_index = node_index;
+
+    // Xoring by 1 switch between 2k <-> 2k + 1, which are siblings in the tree.
+    // The parent of these siblings is k = floor(n/2) for n = 2k, 2k+1.
+    while updated_nodes.contains_key(&(cur_node_index.clone() ^ BigUint::from(1u64))) {
```
notlesh, line 46:
```suggestion
    let one = BigUint::from(1u64);
    while updated_nodes.contains_key(&(cur_node_index.clone() ^ one)) {
```

might be minor, but calling `from()` repeatedly looks like it incurs a heap allocation each time

notlesh (approved):
Awesome work! I left a lot of minor feedback.

src/starkware_utils/commitment_tree/binary_fact_tree_node.rs
odesenfans: Yeah the problem I have is that it's the custom implementation of a method that is then called by a method called `get_children()` with default implementation that relies on `_get_children()`. It's the type of stuff I'd clean up later.

src/starkware_utils/commitment_tree/merkle_tree/traverse_tree.rs
odesenfans: An artefact from experiences with async + recursion + lifetimes, I'll rename it.

src/starkware_utils/commitment_tree/patricia_tree/patricia_utils.rs
odesenfans: I can add a `Default` implementation, but I'll keep this function for symmetry with the Python code.

odesenfans: Nevermind, in this place at least I don't particularly care. This code is only used to compute the root hash using a basic method for use in tests. I don't think it's worth changing anything here.

src/starkware_utils/commitment_tree/patricia_tree/virtual_calculation_node.rs
odesenfans: Thanks, didn't think about that!

src/starkware_utils/commitment_tree/update_tree.rs
odesenfans: At first sight I thought you were right, I tried it but BigUint is not `Copy` so I'd have to clone it each time anyway. Not sure whether the compiler generates different code for multiple calls to `from()` and `clone()` :man_shrugging:

src/starkware_utils/commitment_tree/calculation.rs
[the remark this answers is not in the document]
odesenfans: In order:
* I'll make it `pub(crate)`, sure.
* In terms of documentation, I'll make a separate PR at some point. I'll try to prepare a workshop for next Thursday so this might be part of it.
* I won't rename the trait, I'd like to keep it close to the Python implementation to make it easier to track changes in future versions.

src/starkware_utils/commitment_tree/update_tree.rs
notlesh: You can compare by reference, pass `&one` instead of `one`. Nice!

odesenfans: Good point, that works!

PR #91, opened 2024-03-20
add Cairo1 builtin hint processor

maciejka:
Moves us to the point where Cairo1 contracts are executed to the first syscall:
```
...
executing INVOKE_FUNCTION on: 3221225984
about to call contract_address: 3221225984, class_hash: 2147484160, is_deprecated: 0
TODO: execute syscall hint: SystemCall { system: Deref(CellRef { register: FP, offset: -3 }) }
...
```


src/hints/mod.rs
```diff
@@ -213,22 +225,32 @@ impl HintProcessorLogic for SnosHintProcessor {
...
-        match self.builtin_hint_proc.execute_hint(vm, exec_scopes, hint_data, constants) {
-            Err(HintError::UnknownHint(_)) => {}
-            res => return res.map(|_| HintExtension::default()),
-        }
-        // First attempt to execute with builtin hint processor
-        let hint_data = hint_data.downcast_ref::<HintProcessorData>().ok_or(HintError::WrongHintData)?;
-        let hint_code = hint_data.code.as_str();
-        if let Some(hint_impl) = self.hints.get(hint_code) {
-            return hint_impl(vm, exec_scopes, &hint_data.ids_data, &hint_data.ap_tracking, constants)
+        if let Some(hpd) = hint_data.downcast_ref::<HintProcessorData>() {
```
odesenfans, line 228: Could you put the code in each `if let Some(...) = hint_data.downcast_ref::...` in dedicated functions? It would make this clearer.

odesenfans (approved):
One minor comment, seems good otherwise!

src/hints/mod.rs
maciejka: I will make it in the next pr.

PR #92, opened 2024-03-21
Syscall handler infra

maciejka:
Infrastructure for executing Cairo1 syscalls. Executes first `GetExecutionInfo` syscall, fails on next `CallContract` syscall:
```
...
about to call contract_address: 3221225984, class_hash: 2147484160, is_deprecated: 0
about to execute syscall: GetExecutionInfo
about to execute syscall: CallContract
...
```

A lot of code borrowed from Blockifier. @odesenfans there are a lot of TODOs in syscalls.cairo. This code is adapted from blockifier. I prefer to move it in one go, put not ready parts in comments with todos, then uncomment one by one in subsequent prs.

### DIFF
--- a/src/execution/gas_constants.rs
+++ b/src/execution/gas_constants.rs
@@ -1,0 +1,107 @@
+// Gas Cost.
+// See documentation in core/os/constants.cairo.
+#[allow(unused)]
+pub const STEP_GAS_COST: u64 = 100;
+#[allow(unused)]
+pub const RANGE_CHECK_GAS_COST: u64 = 70;
+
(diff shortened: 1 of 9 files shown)


odesenfans (approved):
Is there no test for any of this code?

maciejka:
> Is there no test for any of this code?

There is an integration test: https://github.com/keep-starknet-strange/snos/blob/29487c6a90cced41eef9b07ace49a4d1d9b74435/tests/os_on_blocks.rs#L38. It fails at some point after executing hints implemented here.

PR #93, opened 2024-03-22
builtins debugging

notlesh:
A dump of debugging code while trying to debug an issue with selected builtins.

I think the culprit is `update_builtin_ptrs` (or some previous logic it depends on) because it looks like it should be finding 2 selected builtins but finds 0, and this leads to a `DiffAssertValue` assertion failure later on.

### DIFF
--- a/src/hints/builtins.rs
+++ b/src/hints/builtins.rs
@@ -8,7 +8,7 @@ use cairo_vm::hint_processor::builtin_hint_processor::hint_utils::{
 use cairo_vm::hint_processor::hint_processor_definition::HintReference;
 use cairo_vm::serde::deserialize_program::ApTracking;
 use cairo_vm::types::exec_scope::ExecutionScopes;
-use cairo_vm::types::relocatable::MaybeRelocatable;
+use cairo_vm::types::relocatable::{MaybeRelocatable, Relocatable};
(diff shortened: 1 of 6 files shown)


src/hints/builtins.rs
```diff
@@ -95,23 +117,45 @@ pub fn update_builtin_ptrs(
...
+                returned_builtins.push(vm.get_maybe(&(selected_ptrs + selected_builtin_offset)?).unwrap());
+                selected_builtin_offset += 1;
+            } else {
+                let val = vm.get_maybe(&(orig_builtin_ptrs + i)?).unwrap();
+                println!("    - adding NON-selected builtin {}", val);
+                let val = if (val == MaybeRelocatable::RelocatableValue(Relocatable{segment_index: -3, offset: 0})) {
+                    MaybeRelocatable::RelocatableValue(Relocatable{segment_index: -3, offset: 1})
+                } else {
+                    val
+                };
```
notlesh, line 147: This ugly hack works around the `DiffAssertValues` issue by modifying a specific pointer when we encounter it.

Remove this in order to reproduce the issue; with it in place we are able to progress further through the OS code.

notlesh:
Closing as #91 solved this, also opening #96 to continue this work.

PR #110, opened 2024-03-28
Feature: SET_SIBLINGS hint

odesenfans:
Issue Number: N/A

## Type

- [x] feature
- [ ] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [ ] yes
- [x] no

### DIFF
--- a/src/hints/patricia.rs
+++ b/src/hints/patricia.rs
@@ -1,0 +1,32 @@
+use std::collections::HashMap;
+
+use cairo_vm::hint_processor::builtin_hint_processor::hint_utils::{get_ptr_from_var_name, insert_value_from_var_name};
+use cairo_vm::hint_processor::hint_processor_definition::HintReference;
+use cairo_vm::serde::deserialize_program::ApTracking;
+use cairo_vm::types::exec_scope::ExecutionScopes;
+use cairo_vm::vm::errors::hint_errors::HintError;
+use cairo_vm::vm::vm_core::VirtualMachine;
+use cairo_vm::Felt252;
+
+use crate::hints::vars;
+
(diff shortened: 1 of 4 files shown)


src/hints/patricia.rs
```diff
@@ -0,0 +1,32 @@
...
+use cairo_vm::types::exec_scope::ExecutionScopes;
+use cairo_vm::vm::errors::hint_errors::HintError;
+use cairo_vm::vm::vm_core::VirtualMachine;
+use cairo_vm::Felt252;
+
+use crate::hints::vars;
+
+pub const SET_SIBLINGS: &str = "memory[ids.siblings], ids.word = descend";
+
+pub type Descend = (Felt252, Felt252);
```
notlesh, line 15: This is not clear to me, why a tuple with two felts? Also, won't this be shared elsewhere (or maybe not outside of hints...)?

notlesh: For that matter, the syntax `memory[ids.siblings], ids.word = descend` isn't clear to me, I interpreted that as:
* set the memory at `ids.siblings` to `descend`
* set `ids.words` to `descend`

You seem to be treating it as a destructuring. If you're clear on this, maybe a comment to help explain it would help.

maciejka (approved):

src/hints/patricia.rs
odesenfans: Regarding the `Descend` type, I just defined it as something that makes sense in this context. I'll refine it once I encounter it in other hints.

For the destructuring/unpacking thing, this is clear Python syntax to me:

```
a, b = x     # a = x[0], b = x[1]
a = b = x   # a = x; b = x
```

PR #114, opened 2024-03-28
Feature: SET_BIT hint

odesenfans:
Issue Number: N/A

## Type

- [x] feature
- [ ] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [ ] yes
- [x] no

### DIFF
--- a/src/hints/patricia.rs
+++ b/src/hints/patricia.rs
@@ -5,11 +5,15 @@ use cairo_vm::hint_processor::builtin_hint_processor::hint_utils::{
 };
 use cairo_vm::hint_processor::hint_processor_definition::HintReference;
 use cairo_vm::serde::deserialize_program::ApTracking;
+use cairo_vm::types::errors::math_errors::MathError;
 use cairo_vm::types::exec_scope::ExecutionScopes;
 use cairo_vm::vm::errors::hint_errors::HintError;
 use cairo_vm::vm::vm_core::VirtualMachine;
 use cairo_vm::Felt252;
+use num_bigint::BigUint;
+use num_traits::ToPrimitive;
 
+use crate::cairo_types::trie::NodeEdge;
 use crate::hints::vars;
 use crate::starkware_utils::commitment_tree::update_tree::DecodeNodeCase;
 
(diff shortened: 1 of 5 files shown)


src/hints/patricia.rs
```diff
@@ -0,0 +1,92 @@
...
+    _exec_scopes: &mut ExecutionScopes,
+    ids_data: &HashMap<String, HintReference>,
+    ap_tracking: &ApTracking,
+    _constants: &HashMap<String, Felt252>,
+) -> Result<(), HintError> {
+    let edge_ptr = get_ptr_from_var_name(vars::ids::EDGE, vm, ids_data, ap_tracking)?;
+    let edge_path = vm.get_integer((edge_ptr + NodeEdge::path_offset())?)?.into_owned();
+    let new_length = {
+        let new_length = get_integer_from_var_name(vars::ids::NEW_LENGTH, vm, ids_data, ap_tracking)?.into_owned();
+        new_length.to_u64().ok_or(MathError::Felt252ToU64Conversion(Box::new(new_length)))?
```
maciejka, line 32: Why is it u64? new_length is in 0..251 range, right?

src/hints/patricia.rs
```diff
@@ -0,0 +1,92 @@
...
+    };
+
+    let bit = (edge_path.to_biguint() >> new_length.to_u64().unwrap()) & BigUint::from(1u64);
```
maciejka, line 35: Does it need to be converted to `u64` twice?

src/hints/patricia.rs
```diff
@@ -0,0 +1,92 @@
...
+        let constants = HashMap::new();
+
+        let ids_data = HashMap::from([
+            (vars::ids::EDGE.to_string(), HintReference::new_simple(-3)),
+            (vars::ids::NEW_LENGTH.to_string(), HintReference::new_simple(-2)),
+            (vars::ids::BIT.to_string(), HintReference::new_simple(-1)),
+        ]);
+
+        let new_length = Felt252::from(3);
+        let edge_path = Felt252::from(0x1008);
```
maciejka, line 74: If `new_length` is 3, then `edge_path <= 2^3`?

src/hints/patricia.rs
odesenfans: `new_length` goes from 0 to 251, not 2^251. Turns out you cannot shift do `BigUint >> BigUint`, but `BigUint >> u64` is supported so that's what I'm doing here.

src/hints/patricia.rs
odesenfans: Yes, you're right about that the test doesn't make a lot of sense with that value. I'll fix that.

maciejka (approved):

PR #117, opened 2024-03-28
Feature: GET_OLD_BLOCK_NUMBER_AND_HASH hint

odesenfans:
Issue Number: N/A

## Type

- [x] feature
- [ ] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [x] yes (one field added to ExecutionHelper)
- [ ] no

### DIFF
--- a/src/execution/helper.rs
+++ b/src/execution/helper.rs
@@ -8,6 +8,7 @@ use blockifier::execution::call_info::CallInfo;
 use blockifier::execution::entry_point_execution::CallResult;
 use blockifier::transaction::objects::TransactionExecutionInfo;
 use cairo_vm::types::relocatable::Relocatable;
+use cairo_vm::vm::errors::hint_errors::HintError;
 use cairo_vm::Felt252;
 use starknet_api::deprecated_contract_class::EntryPointType;
 
@@ -34,6 +35,10 @@ pub struct ExecutionHelper {
     // Pointer to the Cairo ExecutionInfo struct of the current call.
     // Must match the ExecutionInfo pointer for system call validation in 'enter_call'
     pub call_execution_info_ptr: Option<Relocatable>,
+    // The block number and block hash of the (current_block_number - buffer) block, where
+    // buffer=STORED_BLOCK_HASH_BUFFER.
+    // It is the hash that is going to be written by this OS run.
+    pub old_block_number_and_hash: Option<(Felt252, Felt252)>,
     // Iter for CallInfo
     pub call_iter: IntoIter<CallInfo>,
     // CallInfo for the call currently being executed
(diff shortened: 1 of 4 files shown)


src/execution/helper.rs
```diff
@@ -77,6 +83,14 @@ impl ExecutionHelperWrapper {
             })),
         }
     }
+
+    pub fn get_old_block_number_and_hash(&self) -> Result<(Felt252, Felt252), HintError> {
+        let eh_ref = self.execution_helper.as_ref().borrow();
+        eh_ref.old_block_number_and_hash.ok_or(HintError::AssertionFailed(
+            format!("Block number is probably < {STORED_BLOCK_HASH_BUFFER}.").into_boxed_str(),
```
maciejka, line 90: Does this error message make sense?

src/execution/helper.rs
```diff
@@ -34,6 +35,10 @@ pub struct ExecutionHelper {
     // Pointer to the Cairo ExecutionInfo struct of the current call.
     // Must match the ExecutionInfo pointer for system call validation in 'enter_call'
     pub call_execution_info_ptr: Option<Relocatable>,
+    // The block number and block hash of the (current_block_number - buffer) block, where
+    // buffer=STORED_BLOCK_HASH_BUFFER.
+    // It is the hash that is going to be written by this OS run.
+    pub old_block_number_and_hash: Option<(Felt252, Felt252)>,
```
maciejka, line 41: When is it supposed to be `None`?

src/execution/helper.rs
notlesh: The original hint has a pretty specific error message: `"Inconsistent block number. The constant STORED_BLOCK_HASH_BUFFER is probably out of sync."`

maciejka: Here we just unwrap old_block_number_and_hash. If it is None is it an internal error, which has nothing to do with block number being < stored block number.

src/execution/helper.rs
odesenfans: That's a good question, it's just optional in the Python VM so I kept the implementation. Seems weird to me as well.

src/execution/helper.rs
odesenfans: Yeah seems like I skipped a line here. Will fix it.

odesenfans: Nevermind, this is translated directly from `cairo-lang`: 

https://github.com/starkware-libs/cairo-lang/blob/ee7ce74e1159a349d4b77a5f952241b50b1692de/src/starkware/starknet/core/os/syscall_handler.py#L1237

maciejka (approved):

PR #136, opened 2024-04-07
Feature: emit event syscall for Cairo1

maciejka:
Implements Cairo1 `emit event` syscall. Depends on #94 being merged.

PR #137, opened 2024-04-08
Feature: read write syscalls for Cairo1

maciejka:
Depends on #136.


src/hints/execution.rs
```diff
@@ -1469,6 +1469,56 @@ pub fn write_syscall_result(
...
 
+pub const WRITE_SYSCALL_RESULT: &str = indoc! {r#"
+    storage = execution_helper.storage_by_address[ids.contract_address]
+    ids.prev_value = storage.read(key=ids.request.key)
+    storage.write(key=ids.request.key, value=ids.request.value)
+
+    # Fetch a state_entry in this hint and validate it in the update that comes next.
+    ids.state_entry = __dict_manager.get_dict(ids.contract_state_changes)[ids.contract_address]
+    ids.new_state_entry = segments.add()"#
+};
```
odesenfans, line 1480: This hint shares nearly all its implementation with the hint you renamed to `WRITE_SYSCALL_RESULT_DEPRECATED` on line 1421. Could you make them share the same implementation?

maciejka: I don't think it makes sense to invest time into deduplicating code between deprecated and non deprecated versions. At some point deprecated version will be simply removed from the os. Then, it will be easier to have them duplicated. Also it is more possible that they will evolve separately: deprecated version will stay stagnant as it is and non deprecated will be developed.

odesenfans: For me that doesn't matter too much, I see similar code -> it should be deduplicated. But it's not a dealbreaker, let's move on.

odesenfans (approved):

PR #138, opened 2024-04-09
Update unimplemented hints

notlesh:
This PR does a few things:

1. Updates the remaining unimplemented hints.
2. Adds tests to make sure there are no duplicate hint entries
3. Removes duplicate hints
4. Adds some improvements to the hint tool:

    * Moved to its own crate
    * Uses `clap` for CLI args
    * Supports generating rust code, which helps avoid whitespace string matching issues

Example:

image


examples/hints.rs
```diff
@@ -50,6 +50,7 @@ fn main() -> std::io::Result<()> {
         );
         if let Err(e) = r {
             if let HintError::UnknownHint(_) = e {
+                println!(" ----- gave unknown hint error: ----- \n{}", code.clone());
```
odesenfans, line 53: Is this meant to stay?

src/hints/mod.rs
```diff
@@ -224,7 +224,51 @@ static HINTS: [(&str, HintImpl); 166] = [
     (syscalls::STORAGE_WRITE, syscalls::storage_write),
     (syscalls::SET_SYSCALL_PTR, syscalls::set_syscall_ptr),
     (syscalls::VALIDATE_AND_DISCARD_SYSCALL_PTR, syscalls::validate_and_discard_syscall_ptr),
+
+    (unimplemented::maybe_new::UNUSED_HINT_0, hint_stub),
```
odesenfans, line 228: They're not unused, they're unimplemented.

src/hints/mod.rs
```diff
@@ -368,6 +412,17 @@ impl HintProcessorLogic for SnosHintProcessor {
...
 }
 
+pub fn hint_stub(
+    _vm: &mut VirtualMachine,
+    _exec_scopes: &mut ExecutionScopes,
+    _ids_data: &HashMap<String, HintReference>,
+    _ap_tracking: &ApTracking,
+    _constants: &HashMap<String, Felt252>,
+) -> Result<(), HintError> {
+    println!("!!! unimplemented hint !!!");
```
odesenfans, line 422: This should panic.

src/hints/mod.rs
```diff
@@ -224,7 +224,51 @@ static HINTS: [(&str, HintImpl); 166] = [
...
+    (unimplemented::maybe_new::UNUSED_HINT_30, hint_stub),
+    (unimplemented::maybe_new::UNUSED_HINT_31, hint_stub),
+    (unimplemented::maybe_new::UNUSED_HINT_32, hint_stub),
+    (unimplemented::maybe_new::UNUSED_HINT_33, hint_stub),
+    (unimplemented::maybe_new::UNUSED_HINT_34, hint_stub),
+    (unimplemented::maybe_new::UNUSED_HINT_35, hint_stub),
+    (unimplemented::maybe_new::UNUSED_HINT_36, hint_stub),
+    (unimplemented::maybe_new::UNUSED_HINT_37, hint_stub),
+    (unimplemented::maybe_new::UNUSED_HINT_38, hint_stub),
+    (unimplemented::maybe_new::UNUSED_HINT_39, hint_stub),
```
odesenfans, line 267: I'd prefer if you gave them all a name straight away.

odesenfans (changes requested):

src/hints/state.rs
```diff
@@ -137,6 +137,8 @@ pub fn set_preimage_for_current_commitment_info(
     ap_tracking: &ApTracking,
     _constants: &HashMap<String, Felt252>,
 ) -> Result<(), HintError> {
+    let _contract_address = get_integer_from_var_name(vars::ids::CONTRACT_ADDRESS, vm, ids_data, ap_tracking)?.into_owned();
+    // TODO: CommitmentInfo should not be obtained through exec_scopes
```
notlesh, line 141: This was broken previously as it was not using `ids.state_changes.key`

src/hints/mod.rs
notlesh: 8ef3986

examples/hints.rs
notlesh: Removed in 0e995be

src/hints/mod.rs
notlesh: That prevents the hint tool from running (it attempts to call the hint to see if it is implemented)... 

I altered it to return an error in aaecfd3

bin/hint_tool/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+edition = "2021"
+name = "examples"
+version = "0.1.0"
+
+[dependencies]
+blockifier = { git = "https://github.com/keep-starknet-strange/blockifier", branch = "snos/callinfo-clone", features = ["clone", "testing"] }
+cairo-vm = { git = "https://github.com/lambdaclass/cairo-vm", features = ["extensive_hints", "cairo-1-hints"] }
+clap = { version = "4.5.4", features = ["derive"] }
+serde = { version = "1.0.188", features = ["derive"] }
+serde_json = { version = "1.0.105", features = ["arbitrary_precision"] }
```
odesenfans, line 11: Can't you use workspace dependencies for the ones used in the rest of the project? Just to avoid possible conflicts. Not 100% sure how much this makes sense for an example project.

bin/hint_tool/src/main.rs
```diff
@@ -0,0 +1,150 @@
...
+
+#[derive(Clone, Copy, Debug, Default, clap::ValueEnum, PartialEq)]
+enum HintStatus {
+    Implemented,
+    Unimplemented,
+    ImplementedExternally,
+    ImplementedLocally,
+    Orphaned,
+    #[default]
+    All,
```
odesenfans, line 26: What does "All" mean?

src/hints/execution.rs
```diff
@@ -1470,25 +1441,23 @@ pub fn write_syscall_result(
...
+    if tx.version <= 1:
+        assert tx.compiled_class_hash is None, (
+            "Deprecated declare must not have compiled_class_hash."
+        )
+        ids.compiled_class_hash = 0
+    else:
+        assert tx.compiled_class_hash is not None, (
+            "Declare must have a concrete compiled_class_hash."
+        )
+        ids.compiled_class_hash = tx.compiled_class_hash"#
```
odesenfans, line 1457: This type of renaming should be in its own PR.

odesenfans (changes requested):
Some minor comments. I think this should have been split in 2-3 PRs along the way. I need to check the CI and try to run it manually.

bin/hint_tool/src/main.rs
notlesh: This made more sense when this was referred to as "subset" previously. Like you always say, the hardest part is naming :)

Up for suggestions -- `HintStatus` could be renamed (back to `HintSubset` or something like `Filter`)

odesenfans: Filter makes a good deal of sense. Or you could use an `Option<HintStatus>` and disable filtering when it is None.

src/hints/execution.rs
notlesh: Not sure what you mean, the functionality of the hint changed -- it now has nothing to do with any nonce...

odesenfans (approved):

bin/hint_tool/src/main.rs
notlesh: Renamed to `HintFilter` in dbc75df

PR #141, opened 2024-04-10
Convert to Cargo workspace

notlesh:
This PR converts the project to a Cargo workspace structure.

The root `Cargo.toml` is a bit messy as each dependency is duplicated (this isn't exactly obvious from the diff); I think this is a consequence of having a crate in the root directory itself. This could be fixed by moving the `snos` crate to a subdir but that would be very disruptive.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,10 +1,13 @@
-workspace = { members = ["cairo-type-derive", "bin/hint_tool"] }
+[workspace]
+members = ["cairo-type-derive", "bin/hint_tool"]
+resolver = "2"
+
 [package]
 edition = "2021"
 name = "snos"
 version = "0.1.0"
 
-[dependencies]
+[workspace.dependencies]
 anyhow = "1.0.75"
 assert_matches = "1.5.0"
 async-stream = "0.3.5"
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -49,6 +52,51 @@ cairo-lang-semantic = { version = "~2.5.4" }
...
+
+# These dependencies pin cairo-lang requirement versions because cairo-vm does not do it
+cairo-lang-starknet = { workspace = true }
+cairo-lang-casm = { workspace = true }
+cairo-lang-defs = { workspace = true }
+cairo-lang-diagnostics = { workspace = true }
+cairo-lang-filesystem = { workspace = true }
+cairo-lang-semantic = { workspace = true }
+cairo-lang-sierra = { workspace = true }
+cairo-lang-syntax = { workspace = true }
```
odesenfans, line 98: I hadn't thought about this duplication. This limits the interest of the workspace. I'm not sure if this is the right way to go. What do you think?

notlesh: Agreed, it's a nuisance, and isn't very valuable with only a couple minor sub-crates. I'm rather indifferent.

These dependencies would be duplicated in any case, but it would be cleaner if the crate `snos` was in a subdir. We could make that change while we don't have many PRs open... :shrug:

odesenfans:
As decided, let's close this for now.

PR #143, opened 2024-04-10
Impl fp+4 = tx.nonce hint

notlesh:
Implements the `SET_FP_PLUS_4_TO_TX_NONCE` hint.


src/hints/execution.rs
```diff
@@ -1205,6 +1205,22 @@ pub fn set_ap_to_tx_nonce(
...
+pub fn set_fp_plus_4_to_tx_nonce(
+    vm: &mut VirtualMachine,
+    exec_scopes: &mut ExecutionScopes,
+    _ids_data: &HashMap<String, HintReference>,
+    _ap_tracking: &ApTracking,
+    _constants: &HashMap<String, Felt252>,
+) -> Result<(), HintError> {
+    let tx: &InternalTransaction = exec_scopes.get_ref(vars::scopes::TX)?;
+    let nonce = tx.nonce.ok_or(HintError::AssertionFailed("tx.nonce should be set".to_string().into_boxed_str()))?;
+    vm.insert_value((vm.get_fp() + 4)?, nonce).map_err(HintError::Memory)?;
```
odesenfans, line 1219: Why do you `map_err`? `vm.insert_value(...)?` should be enough.

src/hints/execution.rs
```diff
@@ -1871,4 +1887,29 @@ mod tests {
             get_integer_from_var_name(vars::ids::STATE_ENTRY, &mut vm, &ids_data, &ap_tracking).unwrap().into_owned();
         assert_eq!(state_entry, Felt252::from(123));
     }
+
+    #[test]
+    fn test_set_fp_plus_4_to_tx_nonce() {
+        let mut vm = VirtualMachine::new(false);
+        vm.add_memory_segment();
+        vm.add_memory_segment();
+        vm.set_fp(3);
```
odesenfans, line 1896: As you do not insert any variable, you can just leave FP to 0.

odesenfans (changes requested):
Two minor comments.

src/hints/execution.rs
```diff
@@ -1871,4 +1887,29 @@ mod tests {
...
+        let mut tx = InternalTransaction::default();
+        tx.nonce = Some(Felt252::THREE);
+        let mut exec_scopes: ExecutionScopes = Default::default();
+        exec_scopes.insert_value(vars::scopes::TX, tx);
+
+        set_fp_plus_4_to_tx_nonce(&mut vm, &mut exec_scopes, &ids_data, &ap_tracking, &constants).unwrap();
+
+        let address: Relocatable = (vm.get_fp() + 4usize).unwrap();
+        let value = vm.get_integer(address).unwrap();
+        assert_eq!(value, std::borrow::Cow::Borrowed(&Felt252::THREE));
```
maciejka, line 1913: Isn't it simpler to just `assert_eq!(*value, Felt252::THREE)`? Or deref one line above?

src/hints/execution.rs
notlesh: I copied `insert_value_into_ap()` impl verbatim and replaced `ap` with `fp+4`. Seems unnecessary, though -- it's removed in a6f10f8

src/hints/execution.rs
notlesh: Good point. fixed in 0c72f85

src/hints/execution.rs
```diff
@@ -1871,4 +1887,29 @@ mod tests {
             get_integer_from_var_name(vars::ids::STATE_ENTRY, &mut vm, &ids_data, &ap_tracking).unwrap().into_owned();
         assert_eq!(state_entry, Felt252::from(123));
     }
+
+    #[test]
+    fn test_set_fp_plus_4_to_tx_nonce() {
+        let mut vm = VirtualMachine::new(false);
+        vm.add_memory_segment();
```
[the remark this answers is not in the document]
notlesh: I'm not sure, we've been doing this in every test that we've written and I've not questioned it in a long time. Maybe @odesenfans is clear on this...

src/hints/execution.rs
notlesh: Good point, cleaned this up in 45bc33e

src/hints/execution.rs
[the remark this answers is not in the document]
odesenfans: Sorry, missed this comment. This is only useful when declaring Cairo variables, which you're not doing here. So you can remove these segments and leave `fp` to its default value.

odesenfans (approved):

PR #144, opened 2024-04-11
Feature: BUILD_DESCENT_MAP hint

odesenfans:
Issue Number: N/A

## Type

- [x] feature
- [ ] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [ ] yes
- [x] no


maciejka:
I need a walk-through.

src/starkware_utils/commitment_tree/base_types.rs
```diff
@@ -76,3 +89,5 @@ impl Display for Height {
         self.0.fmt(f)
     }
 }
+
+pub type DescentMap = HashMap<(Felt252, Felt252), Vec<Felt252>>;
```
notlesh, line 93: This really needs better typing, and documentation. The only bit I can find about what these types are comes from the comment copied from `cairo-lang`:

```rust
/// Simply put, a descent is a path of length > 1 where there is only one path with data
/// in the tree. Each entry in the descent map is of the form:
/// (start_node_height, start_node_path) -> (descent_length, end_node_relative_path).
/// The key is a tuple representing the absolute position of the node at which the descent starts
/// in the tree. The value represents the descent relative to the start node.
```

Why is this a `Vec<Felt252>` here if it's used consistently (and documented) as a `(Felt, Felt)`? My guess is that reflects the Cairo code and/or pythonic implementation.

IMO it is worth deviating here since we gain so much by having a stronger type system. Most of what is described in the blurb above is defined as strong types in this file already, we have:

```rust
pub struct NodePath(pub BigUint);
pub struct Length(pub u64);
pub struct Height(pub u64);
```

One idea for a stronger typing for `DescentMap`:

```rust
pub struct DescentStart(Height, NodePath);
pub struct DescentPath(Length, NodePath);
pub type DescentMap = HashMap<DescentStart, DescentPath>;
```

The most glaring issue with this is that we're using a mishmash of `u64`s and `BigUint`s now, and it's not clear how annoying this would be for interacting with Cairo.

notlesh (commented):
We really need some docmentation on some of the types here, otherwise it is extremely difficult to understand what a Descent is and what the code or tests are doing with all the tuples and vecs.

I made an argument (see previous comment) about deviating from the pythonic impl in favor of better typing (which is largely self-describing).

src/starkware_utils/commitment_tree/patricia_tree/patricia_guess_descents.rs
```diff
@@ -0,0 +1,475 @@
...
+            DescentMap::from([((Felt252::from(3), Felt252::from(0)), vec![Felt252::from(2), Felt252::from(0)])]),
+        );
+    }
+
+    /// Tests generating a descent map for an update tree of two leaves, against
+    /// an empty tree. The new tree is also empty to have the descent map reflect
+    /// the structure of the update tree and nothing else.
+    #[rstest]
+    fn test_guess_descents_update_two_leaves() {
+        let height = Height(3);
```
notlesh, line 440: The tests only use a height of 3, which I think fails to cover some other important cases. For example, a tree that requires two descents of different lengths.

The "real world" will have a static height of `251` unless I'm missing something. I suggest at least adding a test case with a height on this order to flush out any absurd cases like OOM, etc.

src/starkware_utils/commitment_tree/base_types.rs
notlesh: I made a PR (#162) as a proof of concept, see it here: https://github.com/keep-starknet-strange/snos/pull/162/commits/0387988beacd194036cdfa1a32bedaade85f9040

odesenfans: I agree that it's less than ideal. We can check it out. Btw the comment you mentioned is something I added myself after our discussion last Friday :sweat_smile:  Also, could you open a PR from your branch to mine on the MSL fork? I'd like to see the diff directly.

odesenfans: The idea you propose seems okay to me, let's just check the implications.

src/starkware_utils/commitment_tree/patricia_tree/patricia_guess_descents.rs
odesenfans: Yeah that's probably not a bad idea! I used height 3 because it's easy to reason with.

notlesh (approved):

PR #146, opened 2024-04-11
Feature: ASSIGN_BYTECODE_SEGMENTS hint

odesenfans:
Issue Number: N/A

## Type

- [x] feature
- [ ] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [ ] yes
- [x] no

### DIFF
--- a/src/hints/compiled_class.rs
+++ b/src/hints/compiled_class.rs
@@ -1,0 +1,32 @@
+use std::collections::HashMap;
+
+use cairo_vm::hint_processor::hint_processor_definition::HintReference;
+use cairo_vm::serde::deserialize_program::ApTracking;
+use cairo_vm::types::exec_scope::ExecutionScopes;
+use cairo_vm::vm::errors::hint_errors::HintError;
+use cairo_vm::vm::vm_core::VirtualMachine;
+use cairo_vm::Felt252;
+use indoc::indoc;
+
+use crate::hints::vars;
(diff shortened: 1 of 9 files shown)


maciejka:
Hints related to selective bytecode loading can be similarly to SECP related hints left for later. It is an optimisation that os may or might not do. Not on critical path.

odesenfans:
> Hints related to selective bytecode loading can be similarly to SECP related hints left for later. It is an optimisation that os may or might not do. Not on critical path.

Yes I realized that a bit late. Oh well, it's implemented at least.

notlesh:
> > Looks ok, but something this complex really needs a test. There is at least one more bytecode related hint, btw.
> 
> Is this really complex enough? It's pretty trivial IMO, we just extract one field from a struct and turn it into an iterator. I can add tests if you insist.

The code here isn't, but the overall mechanism is, example: https://github.com/starkware-libs/cairo-lang/blame/master/src/starkware/starknet/core/os/contract_class/compiled_class.cairo#L118

But as @maciejka pointed out, it's not in critical path, so I suggest either close it or merge it for now, up to you.

notlesh (approved):

odesenfans:
> The code here isn't, but the overall mechanism is, example: https://github.com/starkware-libs/cairo-lang/blame/master/src/starkware/starknet/core/os/contract_class/compiled_class.cairo#L118
> 
> But as @maciejka pointed out, it's not in critical path, so I suggest either close it or merge it for now, up to you.

Let's merge it, I don't see a reason not to.

PR #154, opened 2024-04-11
Use `StorageByAddress`

notlesh:
This PR hooks up the `StorageByAddress` map in `ExecutionHelper` which uses the `OsSingleStarknetStorage` and `PatriciaTree` work from @odesenfans. Depends on his #152 because that's where I started diagnosing the issue...

### DIFF
--- a/src/starknet/starknet_storage.rs
+++ b/src/starknet/starknet_storage.rs
@@ -128,6 +128,8 @@ impl CommitmentInfo {
     }
 }
 
+// TODO: document this, is this the entire tree storage for a whole contract?
+//       why is it called "Single"?
 #[derive(Clone, Debug)]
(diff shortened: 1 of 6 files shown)


src/starknet/starknet_storage.rs
```diff
@@ -128,6 +128,8 @@ impl CommitmentInfo {
     }
 }
 
+// TODO: document this, is this the entire tree storage for a whole contract?
+//       why is it called "Single"?
```
notlesh, line 132: TODO: convert this to documentation :)

I'm guessing `Single` mostly comes from the pythonic implementation, but maybe there's a better answer than that.

maciejka:
Make sure `os_on_blocks` tests run up to `state_update` phase after this changes.

PR #158, opened 2024-04-12
Feature: FETCH_RESULT hint

odesenfans:
Issue Number: N/A

## Type

- [x] feature
- [ ] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [ ] yes
- [x] no

### DIFF
--- a/src/hints/execution.rs
+++ b/src/hints/execution.rs
@@ -1726,7 +1726,7 @@ pub fn fetch_result(
     let retdata_size = get_integer_from_var_name(vars::ids::RETDATA_SIZE, vm, ids_data, ap_tracking)?;
 
     // validated is the string "VALID" translated to a felt.
-    let validated = get_constant(vars::constants::VALIDATED, &constants)?;
+    let validated = get_constant(vars::constants::VALIDATED, constants)?;
 
     let n_elements = std::cmp::min(felt_to_usize(&retdata_size)?, 100usize);
 
--- a/src/hints/execution.rs
+++ b/src/hints/execution.rs
@@ -1,4 +1,5 @@
 use std::any::Any;
+use std::borrow::Cow;
 use std::collections::{HashMap, HashSet};
 use std::vec::IntoIter;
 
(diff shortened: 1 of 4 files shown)


src/hints/execution.rs
```diff
@@ -1733,6 +1734,47 @@ pub fn get_old_block_number_and_hash(
...
+    ids_data: &HashMap<String, HintReference>,
+    ap_tracking: &ApTracking,
+    constants: &HashMap<String, Felt252>,
+) -> Result<(), HintError> {
+    // Fetch the result, up to 100 elements.
+    let retdata = get_ptr_from_var_name(vars::ids::RETDATA, vm, ids_data, ap_tracking)?;
+    let retdata_size = get_integer_from_var_name(vars::ids::RETDATA_SIZE, vm, ids_data, ap_tracking)?;
+
+    // validated is the string "VALID" translated to a felt.
+    let validated = constants
```
maciejka, line 1759: Just realized that there is a `get_constant_from_var_name` in the vm.

odesenfans: Missed it as well! I'll replace it.

odesenfans: Oh actually this does not behave as expected. It takes the name of constant (ex: `VALIDATED`) and looks for it through the whole hashmap to find an entry that matches by splitting the path. Very expensive operation that we can entirely avoid. I will add a helper function instead.

src/hints/execution.rs
```diff
@@ -1733,6 +1734,45 @@ pub fn get_old_block_number_and_hash(
...
+    // validated is the string "VALID" translated to a felt.
+    let validated = get_constant_from_var_name(vars::constants::VALIDATED, constants)?;
+
+    let n_elements = std::cmp::min(felt_to_usize(&retdata_size)?, 100usize);
+
+    let result = vm.get_range(retdata, n_elements);
+
+    // This hint is weird, there is absolutely no need to fetch 100 elements to do this.
+    // Nonetheless, we implement it 1-1 with the Python version.
+    if n_elements != 1 || result[0] != Some(Cow::Borrowed(&MaybeRelocatable::Int(*validated))) {
```
notlesh, line 1767: why `if n_elements != 1`?

yes, a weird hint...

src/hints/vars.rs
```diff
@@ -86,4 +86,5 @@ pub mod ids {
 pub mod constants {
     pub const BLOCK_HASH_CONTRACT_ADDRESS: &str = "starkware.starknet.core.os.constants.BLOCK_HASH_CONTRACT_ADDRESS";
     pub const STORED_BLOCK_HASH_BUFFER: &str = "starkware.starknet.core.os.constants.STORED_BLOCK_HASH_BUFFER";
+    pub const VALIDATED: &str = "starkware.starknet.core.os.constants.VALIDATED";
```
notlesh, line 89: This var name is likely to conflict with a future string...

src/hints/execution.rs
```diff
@@ -1733,6 +1734,45 @@ pub fn get_old_block_number_and_hash(
...
+    ids_data: &HashMap<String, HintReference>,
+    ap_tracking: &ApTracking,
+    constants: &HashMap<String, Felt252>,
+) -> Result<(), HintError> {
+    // Fetch the result, up to 100 elements.
+    let retdata = get_ptr_from_var_name(vars::ids::RETDATA, vm, ids_data, ap_tracking)?;
+    let retdata_size = get_integer_from_var_name(vars::ids::RETDATA_SIZE, vm, ids_data, ap_tracking)?;
+
+    // validated is the string "VALID" translated to a felt.
+    let validated = get_constant_from_var_name(vars::constants::VALIDATED, constants)?;
```
notlesh, line 1759: Can you elaborate?

src/hints/execution.rs
odesenfans: Basically it fetches up to 100 elements but hopes that `min(n_elements, 100)` is 1. I don't get why, but maybe this will get clearer later on so I kept the behavior.

src/hints/vars.rs
odesenfans: :man_shrugging:

src/hints/execution.rs
notlesh: I still don't understand the `n_elements != 1` here, it seems like it's a bounds check, but I don't think it works -- if `n_elements == 0` then this will cause an OOB array index with `result[0]`.

If it's a bounds-check, I'd stick with `if results.len() > 0`. Or maybe it's some python that I don't understand...

Also, maybe some context: `__validate__()` is an account abstraction fn that is entirely user-controllable, so maybe this is to mitigate abuse.

notlesh: My second point is wrong, if `n_elements == 1` it should prevent an OOB, but that's relying on behavior of `vm.get_range()`, so my suggestion is still relevant if it's just a bounds check

src/hints/execution.rs
notlesh: Different comment: should this use your `get_constant()` util fn?

src/hints/execution.rs
maciejka: This hints is used in context of transaction validation. It expects result to be [ids.VALIDATED], if it isn't then it prints first 100 elements of the result as an error message.

maciejka (approved):

PR #170, opened 2024-04-16
Fix: share storage across contracts

odesenfans:
Problem: `OsSingleStarknetStorage` is currently built with a different FFC/storage for each contract. This creates N distinct storage objects while we should use one.

Solution: wrap the storage object inside an `Arc<Mutex<_>>` to make it shareable.

Issue Number: N/A

## Type

- [ ] feature
- [ ] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [ ] yes
- [x] no

### DIFF
--- a/src/starkware_utils/commitment_tree/binary_fact_tree_node.rs
+++ b/src/starkware_utils/commitment_tree/binary_fact_tree_node.rs
@@ -1,5 +1,6 @@
 use std::collections::HashMap;
 use std::marker::PhantomData;
+use std::ops::Deref;
 
 use futures::future::FutureExt;
 use num_bigint::BigUint;
@@ -108,9 +109,10 @@ where
 
     async fn get_children(&mut self, node: &Self::NodeType) -> Result<Vec<Self::NodeType>, TreeError> {
         if node.previous.is_leaf() {
+            let storage = self.ffc.storage().await;
             // unwrap() on leaf_hash is guaranteed to be safe because of the check above
-            let previous = LF::get_or_fail(&self.ffc.storage, &node.previous.leaf_hash().unwrap()).await?;
-            let current = LF::get_or_fail(&self.ffc.storage, &node.current.leaf_hash().unwrap()).await?;
+            let previous = LF::get_or_fail(storage.deref(), &node.previous.leaf_hash().unwrap()).await?;
+            let current = LF::get_or_fail(storage.deref(), &node.current.leaf_hash().unwrap()).await?;
             self.result.push(BinaryFactTreeNodeDiff::new(node.path, previous, current));
             return Ok(vec![]);
         }
@@ -258,7 +260,8 @@ where
             // TODO: determine what to do with the assertion in the Python code
             // assert set(indices) == {0}, f"Commitment tree indices out of range: {indices}."
 
-            let leaf = LF::get_or_fail(&ffc.storage, &self._leaf_hash()).await?;
+            let storage = ffc.storage().await;
+            let leaf = LF::get_or_fail(storage.deref(), &self._leaf_hash()).await?;
             Ok(HashMap::from([(BigUint::from(0u64), leaf)]))
         }
         .boxed()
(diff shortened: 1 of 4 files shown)


src/storage/storage.rs
```diff
@@ -154,9 +157,12 @@ where
     S: Storage,
     H: HashFunctionType,
 {
-    #[allow(unused)]
     pub fn new(storage: S) -> Self {
-        Self { storage, _h: Default::default() }
+        Self { storage: Arc::new(Mutex::new(storage)), _h: Default::default() }
+    }
+
+    pub async fn storage(&self) -> tokio::sync::MutexGuard<S> {
```
notlesh, line 164: I'd consider renaming this to convey/warn that it's locking, something like `acquire_storage()` or similar

src/starkware_utils/commitment_tree/binary_fact_tree_node.rs
```diff
@@ -108,9 +109,10 @@ where
 
     async fn get_children(&mut self, node: &Self::NodeType) -> Result<Vec<Self::NodeType>, TreeError> {
         if node.previous.is_leaf() {
+            let storage = self.ffc.storage().await;
             // unwrap() on leaf_hash is guaranteed to be safe because of the check above
-            let previous = LF::get_or_fail(&self.ffc.storage, &node.previous.leaf_hash().unwrap()).await?;
-            let current = LF::get_or_fail(&self.ffc.storage, &node.current.leaf_hash().unwrap()).await?;
+            let previous = LF::get_or_fail(storage.deref(), &node.previous.leaf_hash().unwrap()).await?;
+            let current = LF::get_or_fail(storage.deref(), &node.current.leaf_hash().unwrap()).await?;
```
notlesh, line 115:
```suggestion
            let previous = LF::get_or_fail(&*storage, &node.previous.leaf_hash().unwrap()).await?;
            let current = LF::get_or_fail(&*storage, &node.current.leaf_hash().unwrap()).await?;
```

Maybe, maybe not...

notlesh (commented):
How does this compare to the pythonic implementation? We end up needing this because we followed its concurrency pattern verbatim (right?), and I'm hesitant to deviate from this strategy at the last minute...

odesenfans:
Two parts about this answer.

1. In Python, every object is an `Arc` under the hood.
2. You usually don't need to worry too much about concurrency in Python because of the Global Interpreter Lock (GIL). So the `Mutex` makes sense to add in our context. 
 
I just wanted to see how long we could live without `Arc<Mutex<_>>` in our implementation, turns out that if the storage is shared there's not a lot of other solutions. `Arc<RefCell<_>>` is not enough because the code is async, clippy complains (rightfully so) that holding a mut ref from a refcell across a waiting point will lead to some panicking. So `Arc<Mutex<_>>` it is until we decide we need something more complex.

src/starkware_utils/commitment_tree/binary_fact_tree_node.rs
odesenfans: Yeah the `&*` pattern feels ugly to me, I prefer an explicit `deref()`.

notlesh (approved):
I'm still a bit scared about combining a `mutex` and `await`. Tokio talks a bit about this here, going so far as to suggest avoiding it if it isn't needed: https://docs.rs/tokio/latest/tokio/sync/struct.Mutex.html

If you've considered this carefully, I'll trust your judgement. I'd otherwise suggest using a normal mutex if you can and avoid holding it across `await` points.

Edit: in case it isn't clear, there's `tokio::sync::Mutex` used currently in this PR and there's also `std::sync::Mutex`.

odesenfans:
> I'm still a bit scared about combining a `mutex` and `await`. Tokio talks a bit about this here, going so far as to suggest avoiding it if it isn't needed: https://docs.rs/tokio/latest/tokio/sync/struct.Mutex.html
> 
> If you've considered this carefully, I'll trust your judgement. I'd otherwise suggest using a normal mutex if you can and avoid holding it across `await` points.
> 
> Edit: in case it isn't clear, there's `tokio::sync::Mutex` used currently in this PR and there's also `std::sync::Mutex`.

I understand your concern. I don't see any workaround to holding the resource across waiting points, This PR tries to put it as deep as possible but still the storage methods are asynchronous so you're stuck at some point. Note that this is mostly to make clippy happy: most of the code is async in name only (for now), we perform operations per-contract so there is never any contention on the lock.

Anyway, moving forward we'll probably change the storage into a separate service and use message passing, which seems to be the best pattern available in async Rust.

PR #173, opened 2024-04-16
Impl ASSERT_END_OF_BYTECODE_SEGMENTS hint

notlesh:
Implements the `ASSERT_END_OF_BYTECODE_SEGMENTS` hint.

Issue Number: N/A

## Type

- [x] feature
- [ ] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [ ] yes
- [x] no

### DIFF
--- a/src/hints/compiled_class.rs
+++ b/src/hints/compiled_class.rs
@@ -41,11 +41,11 @@ pub fn assert_end_of_bytecode_segments(
     _ap_tracking: &ApTracking,
     _constants: &HashMap<String, Felt252>,
 ) -> Result<(), HintError> {
-    let mut bytecode_segments: <Vec<BytecodeSegment> as IntoIterator>::IntoIter = exec_scopes.get(vars::scopes::BYTECODE_SEGMENTS)?;
+    let mut bytecode_segments: <Vec<BytecodeSegment> as IntoIterator>::IntoIter =
+        exec_scopes.get(vars::scopes::BYTECODE_SEGMENTS)?;
     if let Some(_) = bytecode_segments.next() {
         return Err(HintError::AssertionFailed("bytecode_segments is not exhausted".to_string().into_boxed_str()));
     }
 
     Ok(())
 }
-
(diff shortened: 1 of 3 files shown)


src/hints/compiled_class.rs
```diff
@@ -30,3 +30,22 @@ pub fn assign_bytecode_segments(
...
+    assert next(bytecode_segments, None) is None"#
+};
+pub fn assert_end_of_bytecode_segments(
+    _vm: &mut VirtualMachine,
+    exec_scopes: &mut ExecutionScopes,
+    _ids_data: &HashMap<String, HintReference>,
+    _ap_tracking: &ApTracking,
+    _constants: &HashMap<String, Felt252>,
+) -> Result<(), HintError> {
+    let mut bytecode_segments: <Vec<BytecodeSegment> as IntoIterator>::IntoIter = exec_scopes.get(vars::scopes::BYTECODE_SEGMENTS)?;
```
notlesh: This type scares me... A test that uses both this hint and the one above (`ASSIGN_BYTECODE_SEGMENTS`) would help.

src/hints/compiled_class.rs
```diff
@@ -30,3 +30,73 @@ pub fn assign_bytecode_segments(
...
+    };
+    use crate::starkware_utils::commitment_tree::base_types::Length;
+
+    #[test]
+    fn test_bytecode_segment_hints() {
+        // tests both ASSIGN_BYTECODE_SEGMENTS and ASSERT_END_OF_BYTECODE_SEGMENTS. The first
+        // should prepare an iterater and put it in ExecutionScopes, the second should ensure that
+        // the iterator is exhausted.
+
+        let mut vm = VirtualMachine::new(false);
```
maciejka, line 69: Default vm setup deserves a fixture. I mean all hint arguments: vm,  exec_scopes, ...

notlesh: I like that idea, especially with the magic values (like needing 2 memory segments and fp to be 2 or more).

By the way, @odesenfans cleared up the two mem segments you asked about earlier:
* Program code executes at segment 0
* Variables are in segment 1
* The VM does not create these segments on its own

I'd support a fixture for this, but I think it should go in a new PR (maybe after we finish up the last few hints).

src/hints/compiled_class.rs
notlesh: Test added in the last few commits

maciejka (approved):

maciejka (commented):
Not on critical path.

odesenfans (approved):

PR #166, opened 2024-04-15
Feature: Impl WRITE_SPLIT_RESULT

notlesh:
Implements the `WRITE_SPLIT_RESULT` hint (renamed from `WRITE_ADDRESS`)

Issue Number: N/A

## Type

- [x] feature
- [ ] bugfix
- [ ] dev (no functional changes, no API changes)
- [ ] fmt (formatting, renaming)
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [ ] yes
- [x] no

PR #175, opened 2024-04-17
Feature: COMPUTE_IDS_LOW hint

odesenfans:
Issue Number: N/A

## Type

- [x] feature
- [ ] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [ ] yes
- [x] no

### DIFF
--- a/src/cairo_types/bigint.rs
+++ b/src/cairo_types/bigint.rs
@@ -1,0 +1,9 @@
+use cairo_type_derive::FieldOffsetGetters;
+use cairo_vm::Felt252;
+
+#[derive(FieldOffsetGetters)]
+pub struct BigInt3 {
+    pub d0: Felt252,
+    pub d1: Felt252,
+    pub d2: Felt252,
+}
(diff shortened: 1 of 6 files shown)


src/cairo_types/bigint.rs
```diff
@@ -0,0 +1,9 @@
+use cairo_type_derive::FieldOffsetGetters;
+use cairo_vm::Felt252;
+
+#[derive(FieldOffsetGetters)]
+pub struct BigInt3 {
+    pub d0: Felt252,
+    pub d1: Felt252,
+    pub d2: Felt252,
+}
```
notlesh, line 9: I avoided adding this in #166 just to reduce code, but I didn't have a strong conviction about doing so... If we're adding it here anyway, that hint could be updated to use it.

notlesh (approved):

src/cairo_types/bigint.rs
odesenfans: Up to you!

PR #177, opened 2024-04-17
Declare V2 test

notlesh:
This PR adds a declare txn v2 test for Cairo1.

Issue Number: N/A

## Type

- [x] feature
- [ ] bugfix
- [x] fmt (formatting, renaming)
- [ ] build
- [ ] docs
- [x] testing

## Breaking changes?

- [ ] yes
- [x] no


tests/integration/common/transaction_utils.rs
```diff
@@ -79,40 +81,155 @@ pub fn to_internal_tx(account_tx: &AccountTransaction) -> InternalTransaction {
...
+    let class_hash: Option<Felt252>;
     let compiled_class_hash: Option<Felt252> = None;
     let calldata: Option<Vec<Felt252>>;
     let paid_on_l1: Option<bool> = None;
-    let r#type: String = "INVOKE_FUNCTION".to_string();
+    let r#type: String;
     let max_fee: Option<Felt252>;
 
     match account_tx {
-        Declare(_) => panic!("Not implemented"),
```
notlesh, line 95: The changes here to `fn to_internal_tx` are pretty messy, I think this will need a refactor in order to handle all txn types we anticipate. Duplicating it for each high-level txn type is probably ok (at the expense of some code duplication).

tests/integration/os.rs
```diff
@@ -91,6 +100,62 @@ async fn return_result_cairo1_account(
...
+    let tx_version = TransactionVersion::TWO;
+    let mut nonce_manager = NonceManager::default();
+
+    let account_contract = initial_state.cairo1_contracts.get("account_with_dummy_validate").unwrap();
+
+    // we want to declare a fresh (never-before-declared) contract, so we don't want to reuse
+    // anything from the test fixtures, and we need to do it "by hand"
+    let (_, sierra_class, casm_class) = load_cairo1_contract("empty_contract");
+    let mut ffc: FactFetchingContext<DictStorage, PoseidonHash> =
+        initial_state.cached_state.state.ffc.clone_with_different_hash();
```
odesenfans, line 122: Nitpick: make an accessor method on `initial state`?

tests/integration/os.rs
```diff
@@ -91,6 +100,62 @@ async fn return_result_cairo1_account(
...
+            sender_address,
+            version: tx_version,
+            nonce: nonce_manager.next(sender_address),
+            class_hash: class_hash.into(),
+            compiled_class_hash,
+        },
+        class_info,
+    );
+
+    initial_state.cairo1_compiled_classes.insert(class_hash, casm_class);
```
odesenfans, line 146: I'm surprised that you have to modify the initial state here. If I understand clearly, in production today, you need to send the Sierra class entirely and Starknet will compile it to CASM. This is obviously not something we're going to support, so I guess the flow here imitates the result of the compilation step?

I think all state manipulations should be isolated and documented to explain why we're doing this as it's not trivial.

tests/integration/os.rs
```diff
@@ -91,6 +100,62 @@ async fn return_result_cairo1_account(
...
+        initial_state.cached_state,
+        block_context,
+        vec![declare_tx],
+        initial_state.cairo0_compiled_classes,
+        initial_state.cairo1_compiled_classes,
+    )
+    .await
+    .expect("OS run failed");
+}
+
```
odesenfans, line 158: Would be cool to have a second test where you deploy the `test_contract` contract and call the `return_result` method.

tests/integration/common/transaction_utils.rs
```diff
@@ -78,47 +80,163 @@ fn tx_hash_invoke_v0(
     ])
 }
 
+fn tx_hash_declare_v1(sender_address: Felt252, max_fee: Felt252, class_hash: Felt252, nonce: Felt252) -> Felt252 {
```
odesenfans, line 83: Document these functions.

odesenfans (changes requested):

tests/integration/common/transaction_utils.rs
```diff
@@ -78,47 +80,163 @@ fn tx_hash_invoke_v0(
...
+// "declare",
+// version,
+// sender_address,
+// 0,
+// h(class_hash),
+// max_fee,
+// chain_id,
+// nonce,
+// compiled_class_hash
+// )
```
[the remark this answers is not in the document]
notlesh: It's a snippet from documentation about the hashing. Removed in c9d3361

tests/integration/os.rs
notlesh: Good catch, this isn't necessary, removed in a6dd142

tests/integration/os.rs
notlesh: Yeah, that was my original goal. I still think we should do that, but I think it should be done as a follow-up.

tests/integration/os.rs
notlesh: Good idea, added in f392552.

tests/integration/common/transaction_utils.rs
notlesh: e0e988a -- pretty basic, let me know if there's something specific you want to see documented

tests/integration/common/transaction_utils.rs
```diff
@@ -78,47 +80,163 @@ fn tx_hash_invoke_v0(
...
+
+                            hash_value = tx_hash_declare_v2(
+                                sender_address_felt,
+                                max_fee.unwrap_or_default(),
+                                felt_api2vm(tx.class_hash.0),
+                                felt_api2vm(tx.compiled_class_hash.0),
+                                nonce_felt,
+                            );
+                        }
+                    }
```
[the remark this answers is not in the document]
notlesh: Ok, I finally tackled this in 0ca3ccf..9d317a2, feels much better now.

tests/integration/common/transaction_utils.rs
notlesh: See https://github.com/keep-starknet-strange/snos/pull/177#discussion_r1627347326

tests/integration/os.rs
```diff
@@ -91,6 +100,62 @@ async fn return_result_cairo1_account(
     .expect("OS run failed");
 }
 
+#[rstest]
+// We need to use the multi_thread runtime to use task::block_in_place for sync -> async calls.
+#[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+async fn declare_cairo1_account(
```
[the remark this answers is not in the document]
notlesh: Done in a731a60. This could use more clean up IMO, but this is a start. Suggestions welcome and/or we can follow up as we add more tests.

tests/integration/declare_txn_tests.rs
```diff
@@ -0,0 +1,68 @@
...
+    let tx_version = TransactionVersion::TWO;
+    let mut nonce_manager = NonceManager::default();
+
+    let account_contract = initial_state.cairo1_contracts.get("account_with_dummy_validate").unwrap();
+
+    // we want to declare a fresh (never-before-declared) contract, so we don't want to reuse
+    // anything from the test fixtures, and we need to do it "by hand"
+    let (_, sierra_class, casm_class) = load_cairo1_contract("empty_contract");
+    let mut ffc = initial_state.clone_ffc::<PoseidonHash>();
+    let (contract_class_hash, compiled_class_hash) =
```
odesenfans, line 36: Still missing a comment here, could be more explicit that we need to store the contract in the storage ahead of the tx.

odesenfans (approved):
One minor comment left, pre-approving.

tests/integration/declare_txn_tests.rs
notlesh: ccd4de9

PR #178, opened 2024-04-18
Tests: reorganize integration tests to avoid dead code warnings

odesenfans:
Problem: code that is not used in all the tests in the `common` module is reported as dead code when compiling.

Solution: move all the tests to a single `integration` directory that will be compiled as one binary, spotting dead code correctly.

Issue Number: N/A

## Type

- [ ] feature
- [x] bugfix
- [ ] build
- [ ] docs
- [x] testing

## Breaking changes?

- [ ] yes
- [x] no


tests/integration/main.rs
```diff
@@ -0,0 +1,4 @@
+mod common;
+mod os;
+mod pie;
+mod sharp;
```
notlesh, line 4: This is just a dummy main? I imagine there's a cleaner way to prevent warnings...

notlesh:
I think this can be done simply with:

`#[cfg(test)]` on the test-only code

odesenfans:
> I think this can be done simply with:
> 
> `#[cfg(test)]` on the test-only code

No, that's not the problem. The code is already test-only code in the `tests` directory. The problem is that easy module at the root of `tests` is compiled as a separate binary. As some of these binaries do not use the test utilities, the compiler marks them as unused when compiling these. There are only two workarounds:

1. Put all the tests in the same binary (as I did here)
2. use `mod common; pub use common::*;" tricks.

As solution 1 also decreases the build time (1 binary -> less time spent on linking) and solution 2 may not warn about truly dead code, I think solution 1 is the way to go.

See https://github.com/rust-lang/rust/issues/46379 for more info.

maciejka:
Is it really the simplest thing that will work?
There is a lot of useless code  in the tests directory. I would clean it up first, hoping to get as a result structure that is simple enough not to need any extra solutions.

maciejka (approved):

PR #179, opened 2024-04-18
Feature: build storage from Blockifier state

odesenfans:
Problem: we currently use the Blockifier state structures inside the OS. We wish to be agnostic of the data source inside the OS.

Solution: Add a mechanism to translate the Blockifier state into an OS-specific storage map at boot.

This fix allows to reach further in the OS execution.

Issue Number: N/A

## Type

- [x] feature
- [ ] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [x] yes
- [ ] no

### DIFF
--- a/src/execution/deprecated_syscall_handler.rs
+++ b/src/execution/deprecated_syscall_handler.rs
@@ -163,7 +163,6 @@ mod test {
     use blockifier::block_context::{BlockContext, FeeTokenAddresses, GasPrices};
     use blockifier::execution::call_info::Retdata;
     use blockifier::execution::entry_point_execution::CallResult;
-    use blockifier::state::cached_state::CachedState;
     use cairo_vm::types::exec_scope::ExecutionScopes;
     use cairo_vm::types::relocatable::{MaybeRelocatable, Relocatable};
     use cairo_vm::vm::vm_core::VirtualMachine;
@@ -176,7 +175,7 @@ mod test {
 
     use crate::config::STORED_BLOCK_HASH_BUFFER;
     use crate::execution::deprecated_syscall_handler::DeprecatedOsSyscallHandlerWrapper;
-    use crate::execution::helper::ExecutionHelperWrapper;
+    use crate::execution::helper::{ContractStorageMap, ExecutionHelperWrapper};
     use crate::hints::vars;
 
     #[fixture]
@@ -216,7 +215,7 @@ mod test {
 
         let execution_infos = Default::default();
         let exec_helper = ExecutionHelperWrapper::new(
-            CachedState::default(),
+            ContractStorageMap::default(),
             execution_infos,
             &block_context,
             old_block_number_and_hash,
--- a/src/execution/helper.rs
+++ b/src/execution/helper.rs
@@ -189,13 +189,6 @@ impl ExecutionHelperWrapper {
         key: Felt252,
         value: Felt252,
     ) -> Result<(), StorageError> {
-        println!(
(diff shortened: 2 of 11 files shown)


notlesh (approved):

tests/common/transaction_utils.rs
```diff
@@ -138,10 +138,9 @@ fn execute_txs(
     txs: Vec<AccountTransaction>,
 ) -> (StarknetOsInput, ExecutionHelperWrapper) {
     let internal_txs: Vec<_> = txs.iter().map(to_internal_tx).collect();
-    let initial_state = copy_state(&state);
     let execution_infos =
         txs.into_iter().map(|tx| tx.execute(&mut state, block_context, true, true).unwrap()).collect();
-    os_hints(&block_context, initial_state, internal_txs, execution_infos)
+    os_hints(&block_context, state, internal_txs, execution_infos)
```
maciejka, line 143: Both pre-state and post-state is necessary here. See further comments.

tests/common/block_utils.rs
```diff
@@ -162,41 +163,19 @@ pub fn test_state(
...
-
 pub fn os_hints(
     block_context: &BlockContext,
-    mut state: CachedState<DictStateReader>,
+    mut blockifier_state: CachedState<DictStateReader>,
     transactions: Vec<InternalTransaction>,
     tx_execution_infos: Vec<TransactionExecutionInfo>,
 ) -> (StarknetOsInput, ExecutionHelperWrapper) {
-    let mut contracts: HashMap<Felt252, ContractState> = state
+    let mut contracts: HashMap<Felt252, ContractState> = blockifier_state
```
maciejka, line 172: A comment: here we need a post-state, including freshly deployed contracts.

tests/common/block_utils.rs
```diff
@@ -287,8 +266,11 @@ pub fn os_hints(
         block_hash: Default::default(),
     };
 
+    // Convert the Blockifier storage into an OS-compatible one
+    let contract_storage_map = build_starknet_storage(&mut blockifier_state);
```
maciejka, line 270: Here we need a pre-state as `contract_storage_map` will be used for storage syscalls execution.
See, e.g. https://github.com/starkware-libs/cairo-lang/blob/efa9648f57568aad8f8a13fbf027d2de7c63c2c0/src/starkware/starknet/core/os/execution/execute_syscalls.cairo#L670

maciejka (changes requested):

tests/common/block_utils.rs
odesenfans: I'm using the property that `CachedStorage<DictStateReader>` contains both the initial and final states. The initial state remains available through `storage_view` while the final state can be obtained by using `to_state_diff()`.

maciejka: Now I see it.

tests/common/block_utils.rs
```diff
@@ -287,8 +266,11 @@ pub fn os_hints(
...
+
     let execution_helper = ExecutionHelperWrapper::new(
-        state,
+        contract_storage_map,
```
maciejka, line 273: Now, it seems like we have two similar contract storage related structures: `contract_storage_map` in `execution_helper` and `contracts` in  `os_input`. Do we need both or is it just temporary?

odesenfans: I see the same behaviour in the Python VM structures so I'd say this is fine, at least for now.

maciejka (approved):
LGTM

PR #182, opened 2024-04-18
Create storage for final storage if needed

notlesh:
Previously we created `OsSingleStarknetStorage` for the initial state from Blockifier, but this caused problems when this didn't include contracts that were going to be written to for the first time in a block. So we now use contracts from the superset of {initial, final} state.

This also changes the write syscall so that it tolerates missing `prev_value`. I'm not sure what the correct behavior is here, but returning `default` in this case seems to work.

Issue Number: N/A

## Type

- [ ] feature
- [ ] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [ ] yes
- [ ] no


notlesh:
Replacing with https://github.com/Moonsong-Labs/snos/pull/15

PR #183, opened 2024-04-21
Feature: handle cairo1 `send_message_to_l1` syscall

maciejka:
Issue Number: N/A

## Type

- [x] feature
- [ ] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [ ] yes
- [ ] no

### DIFF
--- a/tests/os.rs
+++ b/tests/os.rs
@@ -124,7 +124,21 @@ fn syscalls_cairo1(block_context: BlockContext, initial_state: InitialState, max
         nonce: nonce_manager.next(sender_address),
     });
 
-    let txs = vec![test_emit_event_tx, test_storage_read_write_tx, test_get_block_hash_tx];
+    // test_send_message_to_l1
+
+    let to_address = stark_felt!(1234_u16);
+    let payload = vec![stark_felt!(2019_u16), stark_felt!(2020_u16), stark_felt!(2021_u16)];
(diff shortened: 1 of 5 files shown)


tests/os.rs
```diff
@@ -124,7 +124,21 @@ fn syscalls_cairo1(block_context: BlockContext, initial_state: InitialState, max
...
+
+    let test_send_message_to_l1_tx = test_utils::account_invoke_tx(invoke_tx_args! {
+        max_fee,
+        sender_address: sender_address,
+        calldata: create_calldata(contract_address, "test_send_message_to_l1", entrypoint_args),
+        version: tx_version,
+        nonce: nonce_manager.next(sender_address),
+    });
+
+    let txs = vec![test_emit_event_tx, test_storage_read_write_tx, test_get_block_hash_tx, test_send_message_to_l1_tx];
```
odesenfans, line 141: Unsure of how these tests work, but shouldn't there be at least an assertion somewhere to test that the code works?

odesenfans (approved):
Just one minor question, can be answered async.

tests/os.rs
maciejka: You mean that each individual transaction succeeded?

odesenfans: Yes, and this does not not make sense in an integration test, thanks for the clarification.

PR #194, opened 2024-04-23
Implement cairo1 deploy syscall

maciejka:
Implements cairo1 deploy syscall. It is interesting because it uses state_diff to prepare os_hints data.

### DIFF
--- a/src/execution/syscalls.rs
+++ b/src/execution/syscalls.rs
@@ -101,10 +101,10 @@ pub fn deploy(
     exec_wrapper: &mut ExecutionHelperWrapper,
     remaining_gas: &mut u64,
 ) -> SyscallResult<DeployResponse> {
-
     let execution_helper = &mut exec_wrapper.execution_helper.as_ref().borrow_mut();
 
-    let result = execution_helper.result_iter
+    let result = execution_helper
+        .result_iter
         .next()
         .ok_or(SyscallExecutionError::InternalError(Box::from("No result left in the result iterator.")))?;
 
@@ -121,11 +121,9 @@ pub fn deploy(
 
     let constructor_retdata = ReadOnlySegment { start_ptr, length: retdata.len() };
 
-    let contract_address = execution_helper.deployed_contracts_iter.next().ok_or(
-        HintError::SyscallError("n: No more deployed contracts available to replay".to_string().into_boxed_str()),
-    )?;
-
-    println!("new contract_address: {}", contract_address);
+    let contract_address = execution_helper.deployed_contracts_iter.next().ok_or(HintError::SyscallError(
+        "n: No more deployed contracts available to replay".to_string().into_boxed_str(),
+    ))?;
 
     Ok(DeployResponse { contract_address, constructor_retdata })
 }
--- a/src/execution/syscalls.rs
+++ b/src/execution/syscalls.rs
@@ -63,12 +63,7 @@ pub fn call_contract(
 }
 
 #[derive(Debug, Eq, PartialEq)]
-pub struct DeployRequest {
-    // pub class_hash: Felt252,
-    // pub contract_address_salt: Felt252,
-    // pub constructor_calldata: Vec<Felt252>,
-    // pub deploy_from_zero: bool,
-}
+pub struct DeployRequest {}
 
 impl SyscallRequest for DeployRequest {
     fn read(_vm: &VirtualMachine, ptr: &mut Relocatable) -> SyscallResult<DeployRequest> {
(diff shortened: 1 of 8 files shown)


notlesh:
FYI, I have some work on Cairo0 and Cairo1 declare tests in #177

src/execution/syscalls.rs
```diff
@@ -62,56 +62,66 @@ pub fn call_contract(
...
+    let execution_helper = &mut exec_wrapper.execution_helper.as_ref().borrow_mut();
+
+    let result = execution_helper
+        .result_iter
+        .next()
+        .ok_or(SyscallExecutionError::InternalError(Box::from("No result left in the result iterator.")))?;
+
+    *remaining_gas -= result.gas_consumed;
+
+    let retdata = result.retdata.0.iter().map(|sf| felt_api2vm(*sf)).collect();
```
odesenfans, line 108: IMO it's surprising to see `felt_api2vm` inside the OS implementation, I'd say these conversions should be done before. But that's not the topic of this PR so okay.

src/execution/syscalls.rs
```diff
@@ -62,56 +62,66 @@ pub fn call_contract(
...
+        Ok(())
+    }
+}
+
+pub fn deploy(
+    _request: DeployRequest,
+    vm: &mut VirtualMachine,
+    exec_wrapper: &mut ExecutionHelperWrapper,
+    remaining_gas: &mut u64,
+) -> SyscallResult<DeployResponse> {
```
odesenfans, line 98: A little documentation of what this syscall does would be helpful.

tests/integration/common/block_utils.rs
```diff
@@ -167,13 +167,23 @@ pub fn os_hints(
...
-    let mut contracts: HashMap<Felt252, ContractState> = blockifier_state
-        .state
-        .address_to_class_hash
-        .keys()
+    let state_diff = blockifier_state.to_state_diff();
+    let deployed_addresses = state_diff.address_to_class_hash.keys().cloned().collect::<HashSet<_>>();
+
+    let initial_addresses = blockifier_state.state.address_to_class_hash.keys().cloned().collect::<HashSet<_>>();
+
+    let addresses = initial_addresses.union(&deployed_addresses);
```
odesenfans, line 175: The way you compute this is inefficient because you allocate 3 hashsets instead of one. You can do the following instead:

```
let addresses: HashSet<_> = deployed_addresses.keys().zip(initial_addresses.keys()).collect();
```

As you iterate over this again just below, you can even not collect() the iterator at all and just use it directly instead of `addresses.into_iter()`.

tests/integration/common/block_utils.rs
```diff
@@ -167,13 +167,23 @@ pub fn os_hints(
...
+
+    let mut contracts: HashMap<Felt252, ContractState> = addresses
+        .into_iter()
         .map(|address| {
+            let contract_hash = if deployed_addresses.contains(address) {
+                Felt252::ZERO
```
odesenfans, line 181: Why do you set it by default? Is it because it doesn't exist? Shouldn't this be an `Option` type instead?

tests/integration/common/block_utils.rs
```diff
@@ -167,13 +167,23 @@ pub fn os_hints(
...
+            } else {
+                to_felt252(&blockifier_state.get_class_hash_at(*address).unwrap().0)
+            };
             let contract_state = ContractState {
-                contract_hash: to_felt252(&blockifier_state.state.address_to_class_hash.get(address).unwrap().0),
+                contract_hash,
```
odesenfans, line 186: A generic comment over this block of code: I'd prefer if it were isolated in a function. It's not super clear what it does, isolating it + adding doc would help.

src/execution/syscalls.rs
maciejka: Agree. It will go away once we have an abstraction over blockifier types.

tests/integration/common/block_utils.rs
maciejka: It is exepected by the cairo code: https://github.com/starkware-libs/cairo-lang/blob/efa9648f57568aad8f8a13fbf027d2de7c63c2c0/src/starkware/starknet/core/os/execution/deprecated_execute_syscalls.cairo#L742

src/execution/syscalls.rs
odesenfans: I mean, what the implementation does to implement this syscall. It's not immediately clear from the code.

tests/integration/common/block_utils.rs
maciejka: Added a comment.

tests/integration/common/block_utils.rs
maciejka: you mean contract_hash calculation?

odesenfans: The whole 20-ish lines of code. But anyway, I'll probably wrap this differently with the coming PR so it's not a problem.

PR #196, opened 2024-04-24
Move block number mapping update out of `test_state`

maciejka:
Moves block number mapping update to `execute_txs` out of `test_state`.


notlesh (approved):
LGTM. One suggestion: make ` let mut state = CachedState...` not mutable in `test_state()`.

maciejka:
> LGTM. One suggestion: make ` let mut state = CachedState...` not mutable in `test_state()`.

`fund_account` needs `&mut state`.

PR #197, opened 2024-04-26
Cairo1 syscalls refactoring

maciejka:
Combines `Request`, `Response`, `syscall_exec` into a single trait which makes code more straight forward.


src/cairo_types/mod.rs
```diff
@@ -1,6 +1,7 @@
 pub(crate) mod bigint;
 pub(crate) mod builtins;
 pub(crate) mod dict_access;
+pub(crate) mod new_syscalls;
```
notlesh, line 4: "new" isn't very clear to me. suggestions:

* `cairo1_syscalls` (may not be strictly correct)
* `current_syscalls`
* a brief explanation inside `new_syscalls.rs` about new vs. deprecated

maciejka: There is a corresponding new_syscalls.cairo file.

src/execution/syscall_handler.rs
```diff
@@ -59,47 +61,619 @@ impl OsSyscallHandlerWrapper {
...
+    fn write_response(
+        _response: Self::Response,
+        _vm: &mut VirtualMachine,
+        _ptr: &mut Relocatable,
+    ) -> WriteResponseResult {
+        Ok(())
+    }
+}
+
+// TODO: Keccak syscall.
```
notlesh, line 467: Just to confirm: all of this commented out code was just moved, right?

src/execution/syscall_handler_utils.rs
```diff
@@ -300,54 +222,83 @@ pub fn write_segment(vm: &mut VirtualMachine, ptr: &mut Relocatable, segment: Re
...
 }
 
-impl SyscallResponse for SingleSegmentResponse {
-    fn write(self, vm: &mut VirtualMachine, ptr: &mut Relocatable) -> WriteResponseResult {
-        write_segment(vm, ptr, self.segment)
-    }
+pub trait SyscallHandler {
+    type Request;
+    type Response;
+    fn read_request(vm: &VirtualMachine, ptr: &mut Relocatable) -> SyscallResult<Self::Request>;
```
notlesh, line 228: This is already an improvement, but I think we could also generate the `read_request()` and `write_response()` code. WDYT?

notlesh (approved):

src/execution/syscall_handler_utils.rs
maciejka: you mean noop default implementation?

PR #198, opened 2024-04-29
Update test expectations

notlesh:
This PR updates the `StarknetGeneralConfig` struct to reflect upstream changes (namely, removing `cairo_resource_fee_weights` as it was removed upstream.)

It also fixes an intermittent(-ish) test failure by changing the `Default` impl for `StarknetGeneralConfig` so that its behavior does not depend on the existence/contents of a default config file. This was causing the `parse_starknet_config` test to fail in some cases.

Instead, `Default` returns fixed values and a dedicated `fn from_default_file()` is provided (and used in the test).

Issue Number: N/A

## Type

- [ ] feature
- [x] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [ ] yes
- [ ] no

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -82,31 +82,28 @@ pub struct StarknetGeneralConfig {
 
 impl Default for StarknetGeneralConfig {
     fn default() -> Self {
-        match StarknetGeneralConfig::from_file(PathBuf::from(DEFAULT_CONFIG_PATH)) {
-            Ok(conf) => conf,
-            Err(_) => Self {
-                starknet_os_config: StarknetOsConfig {
-                    chain_id: ChainId(SN_GOERLI.to_string()),
-                    fee_token_address: contract_address!(DEFAULT_FEE_TOKEN_ADDR),
-                    deprecated_fee_token_address: contract_address!(DEFAULT_DEPRECATED_FEE_TOKEN_ADDR),
-                },
-                gas_price_bounds: GasPriceBounds {
-                    max_fri_l1_data_gas_price: 10000000000,
-                    max_fri_l1_gas_price: 100000000000000,
-                    min_fri_l1_data_gas_price: 10,
-                    min_fri_l1_gas_price: 100000000000,
-                    min_wei_l1_data_gas_price: 100000,
-                    min_wei_l1_gas_price: 10000000000,
-                },
-                invoke_tx_max_n_steps: MAX_STEPS_PER_TX as u32,
-                validate_max_n_steps: MAX_STEPS_PER_TX as u32,
-                default_eth_price_in_fri: 1_000_000_000_000_000_000_000,
-                constant_gas_price: false,
-                sequencer_address: contract_address!(SEQUENCER_ADDR_0_13_0),
-                cairo_resource_fee_weights: Arc::new(HashMap::from([(N_STEPS_RESOURCE.to_string(), 1.0)])),
-                enforce_l1_handler_fee: true,
-                use_kzg_da: false,
+        Self {
+            starknet_os_config: StarknetOsConfig {
+                chain_id: ChainId(SN_GOERLI.to_string()),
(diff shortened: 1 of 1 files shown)


src/config.rs
```diff
@@ -101,7 +100,6 @@ impl Default for StarknetGeneralConfig {
             default_eth_price_in_fri: 1_000_000_000_000_000_000_000,
             constant_gas_price: false,
             sequencer_address: contract_address!(SEQUENCER_ADDR_0_13_0),
-            cairo_resource_fee_weights: Arc::new(HashMap::from([(N_STEPS_RESOURCE.to_string(), 1.0)])),
```
notlesh: This was removed in `cairo-lang`: https://github.com/starkware-libs/cairo-lang/commit/efa9648f57568aad8f8a13fbf027d2de7c63c2c0

src/config.rs
```diff
@@ -126,7 +123,7 @@ impl StarknetGeneralConfig {
                 eth_fee_token_address: self.starknet_os_config.fee_token_address,
                 strk_fee_token_address: contract_address!("0x0"),
             },
-            vm_resource_fee_cost: self.cairo_resource_fee_weights.clone(),
+            vm_resource_fee_cost: Default::default(), // TODO
```
odesenfans, line 126: Why did you add this TODO?

src/config.rs
```diff
@@ -163,14 +159,14 @@ mod tests {
...
         let expected_seq_addr = contract_address!(SEQUENCER_ADDR_0_13_0);
 
-        let conf = StarknetGeneralConfig::default();
+        let conf = StarknetGeneralConfig::from_default_file().expect("Failed to load default config file");
 
         assert!(!conf.constant_gas_price);
         assert!(conf.enforce_l1_handler_fee);
 
-        assert_eq!(4000000, conf.invoke_tx_max_n_steps);
+        assert_eq!(1000000, conf.invoke_tx_max_n_steps);
```
odesenfans, line 167: Where did that value change?

notlesh: It's defined two ways:
* in `cairo-lang` it comes from `general_config.yml` where it's set to `1_000_000`
* in `blockifier` as `pub const MAX_STEPS_PER_TX: usize = 4_000_000;`

I don't think these values have changed recently.

However, the previous behavior of `Default` for `StarknetGeneralConfig` was done in such a way that it tried to pull values from the file `./cairo-lang/src/starkware/starknet/definitions/general_config.yml`. If that failed (either because it didn't exist or threw a parse error), it would fall back to hard-coded values. In this case, `4000000` was the hard-coded value (`MAX_STEPS_PER_TX` in `blockifier`) and `1000000` comes from a successful parse of `cairo-lang`'s config file.

Essentially, this behavior was silently masking a parse error (`cairo_resource_fee_weights` was removed) and giving us static values instead.

src/config.rs
notlesh: `cairo_resource_fee_weights` was removed upstream, but `vm_resource_fee_cost` still exists. It's not clear what we should do here...

`cairo_lang` has a `versioned_constants.json` with this definition:
```json
    "vm_resource_fee_cost": {
        "bitwise_builtin": 0.16,
        "ec_op_builtin": 2.56,
        "ecdsa_builtin": 5.12,
        "keccak_builtin": 5.12,
        "n_steps": 0.0025,
        "output_builtin": 0,
        "pedersen_builtin": 0.08,
        "poseidon_builtin": 0.08,
        "range_check_builtin": 0.04
    }
```

The previous values came from `general_config.yml`, which was:
```yml
cairo_resource_fee_weights:
  n_steps: 0.0025
```

removed here: https://github.com/starkware-libs/cairo-lang/commit/efa9648f57568aad8f8a13fbf027d2de7c63c2c0#diff-963ec2e2078343b8f2b0a4f018c8091ab7227b1e32972e4d123621a18ed52f47

odesenfans (approved):
One minor question left, approving already.

PR #207, opened 2024-05-20
Remove FeatureContracts

notlesh:
This PR removes the use of `FeatureContract`s in our test code. The main problem with these was their use of faked class hashes.

The PR also cleans up some of the test fixtures and the structs used in test setup and execution.

## Type

- [X] feature
- [ ] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [ ] yes
- [X] no


tests/integration/common/state.rs
```diff
@@ -1,49 +1,92 @@
...
 #[fixture]
-pub fn initial_state(block_context: BlockContext) -> InitialState {
-    let account_without_validations_cairo1 = FeatureContract::AccountWithoutValidations(CairoVersion::Cairo1);
-    let account_without_validations_cairo0 = FeatureContract::AccountWithoutValidations(CairoVersion::Cairo0);
-    let test_contract_cairo1 = FeatureContract::TestContract(CairoVersion::Cairo1);
-    let test_contract_cairo0 = FeatureContract::TestContract(CairoVersion::Cairo0);
-    let erc20_cairo0 = FeatureContract::ERC20;
-
-    let state = test_state(
+pub async fn initial_state(block_context: BlockContext) -> TestState {
```
odesenfans, line 72: What do you think about having a Cairo 0 specific state? I think it would be better to isolate it so that if we have a bug in Cairo1-related code we can more easily find the issue. We can disregard the ERC20 contract being Cairo 0 for now (the Cairo1 test will have 1 Cairo0 contract so it's not fully independent, that's life) but at least the Cairo 0 test is 100% Cairo 0.

odesenfans (approved):
One comment on Cairo 0 vs Cairo 1 code but we can address this in a separate PR.

PR #215, opened 2024-05-30
Internal: introduce the `Hash` type for hashes

odesenfans:
Problem: we did not use a custom type when porting most of the code, resulting in hashes being `Vec<u8>` all over the place. This is inconvenient and error-prone.

Solution: introduce the `Hash` type to encapsulate all hashes. With this, hashes are now type-safe and provide implementations of `From` for common types in Cairo VM and Starknet API.

Issue Number: N/A

## Type

- [ ] feature
- [ ] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [ ] yes
- [x] no

### DIFF
--- a/src/storage/storage.rs
+++ b/src/storage/storage.rs
@@ -1,14 +1,20 @@
 use std::collections::HashMap;
 use std::marker::PhantomData;
-use std::ops::DerefMut;
+use std::ops::{Deref, DerefMut};
 use std::sync::Arc;
 
 use async_stream::try_stream;
 use blockifier::state::errors::StateError;
 use cairo_vm::Felt252;
+use num_bigint::BigUint;
+use serde::{Deserialize, Serialize};
+use starknet_api::core::{ClassHash, CompiledClassHash};
+use starknet_api::hash::StarkFelt;
+use starknet_api::StarknetApiError;
 use tokio::sync::Mutex;
 use tokio_stream::Stream;
 
+use crate::starkware_utils::commitment_tree::patricia_tree::patricia_tree::EMPTY_NODE_HASH;
 use crate::starkware_utils::serializable::{DeserializeError, Serializable, SerializeError};
 
 pub const HASH_BYTES: usize = 32;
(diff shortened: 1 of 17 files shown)


src/storage/storage.rs
```diff
@@ -87,12 +93,106 @@ pub trait Storage: Sync + Send {
...
+impl Hash {
+    pub fn empty() -> Self {
+        Self::from_bytes_be(EMPTY_NODE_HASH)
+    }
+
+    pub fn from_bytes_be(bytes: [u8; 32]) -> Self {
+        Self(bytes)
+    }
+
+    pub fn from_bytes_be_slice(bytes: &[u8]) -> Self {
```
notlesh, line 112: This could use some documentation about behavior WRT input slice length. It looks to me like the only sane way to use it would be to pass an array of exactly 32 bytes, so maybe it needs an assertion about this?

src/storage/storage.rs
```diff
@@ -87,12 +93,106 @@ pub trait Storage: Sync + Send {
     }
 }
 
+/// Starknet hash type.
+/// Encapsulates the result of hash functions and provides conversion functions to Cairo VM
+/// and Starknet API types for convenience.
+#[derive(Debug, Clone, Hash, Eq, PartialEq, Serialize, Deserialize)]
+#[serde(transparent)]
+pub struct Hash([u8; 32]);
```
notlesh, line 101: `Hash` as a name leaves a lot to the imagination...

src/storage/storage.rs
odesenfans: Yeah it could use more doc. Only sane way is if the input is 32 bytes indeed, I implemented it in a way that it supports smaller (add padding) and larger inputs (ignore everything after byte 32).

src/storage/storage.rs
odesenfans: I know :cry: Any idea for a better name? `StarkHash` is taken.

notlesh: In 4 days I failed to come up with a suggestion, so let's live with `Hash` for now :shrug:

notlesh (approved):

PR #217, opened 2024-06-03
Feature: enable class tree updates

odesenfans:
Problem: we still pass an empty commitment info to the OS at startup.

Solution: use the correct class commitment info generated from the SharedState objects.

Additionally:
* Implement the `use_kzg_da` hint, note that the implementation only returns `false` for now.
* Added a mechanism to switch between Pedersen and Poseidon hashing in the Patricia hints. We enable a boolean flag after finishing the state update.

The tests now fail when building the OS output.

Issue Number: N/A

## Type

- [x] feature
- [ ] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [ ] yes
- [x] no

### DIFF
--- a/src/hints/state.rs
+++ b/src/hints/state.rs
@@ -14,11 +14,10 @@ use cairo_vm::{any_box, Felt252};
 use indoc::indoc;
 
 use super::bls_utils::split;
-use crate::cairo_types::builtins::{HashBuiltin, SpongeHashBuiltin};
 use crate::cairo_types::traits::CairoType;
 use crate::cairo_types::trie::NodeEdge;
 use crate::execution::helper::ExecutionHelperWrapper;
-use crate::hints::types::{skip_verification_if_configured, Preimage};
+use crate::hints::types::{get_hash_builtin_fields, skip_verification_if_configured, PatriciaTreeMode, Preimage};
 use crate::hints::vars;
 use crate::io::input::StarknetOsInput;
 use crate::starknet::starknet_storage::{CommitmentInfo, StorageLeaf};
@@ -112,7 +111,7 @@ pub fn set_preimage_for_class_commitments(
     )?;
 
     log::debug!("Setting class trie mode");
-    exec_scopes.data[0].insert(vars::scopes::CLASS_TRIE_MODE.to_string(), any_box!(true));
+    exec_scopes.data[0].insert(vars::scopes::PATRICIA_TREE_MODE.to_string(), any_box!(true));
 
     let preimage = os_input.contract_class_commitment_info.commitment_facts;
     exec_scopes.insert_value(vars::scopes::PREIMAGE, preimage);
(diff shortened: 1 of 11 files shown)


src/hints/state.rs
```diff
@@ -111,6 +110,9 @@ pub fn set_preimage_for_class_commitments(
         ap_tracking,
     )?;
 
+    log::debug!("Setting class trie mode");
+    exec_scopes.data[0].insert(vars::scopes::PATRICIA_TREE_MODE.to_string(), any_box!(PatriciaTreeMode::Class));
```
notlesh, line 114: Do you expect this to do The Right Thing with the nesting of exec scopes?

src/hints/types.rs
```diff
@@ -14,6 +16,32 @@ pub struct PatriciaSkipValidationRunner {
...
+/// Returns the offsets of the x, y and result fields of the hash struct used during Patricia
+/// tree updates depending on the tree being updated.
+pub fn get_hash_builtin_fields(exec_scopes: &ExecutionScopes) -> Result<(usize, usize, usize), HintError> {
+    let patricia_tree_mode: PatriciaTreeMode =
+        get_variable_from_root_exec_scope(exec_scopes, vars::scopes::PATRICIA_TREE_MODE)?;
+    log::debug!("Patricia tree mode: {patricia_tree_mode:?}");
+    Ok(match patricia_tree_mode {
+        PatriciaTreeMode::Class => {
+            (SpongeHashBuiltin::x_offset(), SpongeHashBuiltin::y_offset(), SpongeHashBuiltin::result_offset())
+        }
```
notlesh, line 40:
```suggestion
        PatriciaTreeMode::Class => (SpongeHashBuiltin::x_offset(), SpongeHashBuiltin::y_offset(), SpongeHashBuiltin::result_offset())
```

notlesh (approved):
:rocket: 

Left some minor comments

src/hints/state.rs
odesenfans: Yes, the point is actually to store stuff like that in the root-level scope so that we don't need to worry about nesting.

src/hints/types.rs
odesenfans: What's the difference?

tests/integration/common/block_utils.rs
```diff
@@ -304,11 +304,34 @@ pub async fn os_hints(
...
         )
         .await
-        .expect("Could not create contract state commitment info");
+        .unwrap_or_else(|e| panic!("Could not create contract state commitment info: {:?}", e));
+
+    let accessed_contracts: Vec<TreeIndex> = state_diff
+        .class_hash_to_compiled_class_hash
+        .keys()
+        .chain(compiled_classes.keys())
+        // .chain(deprecated_compiled_classes.keys())
```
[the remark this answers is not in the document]
odesenfans: Not sure we need them, no. Let's see what happens with declare txs for Cairo 0.

PR #218, opened 2024-06-03
Fix: build Starknet OS output struct from VM output

odesenfans:
Problem: the integration tests failed because the code to build the Starknet OS structure had never been used with real outputs.

Solution: reimplement the `from_run` method of `StarknetOsOutput`. The new version also supports KZG DA.

Issue Number: N/A

## Type

- [x] feature
- [x] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [ ] yes
- [x] no

### DIFF
--- a/src/io/output.rs
+++ b/src/io/output.rs
@@ -49,7 +49,8 @@ impl StarknetOsOutput {
 /// Gets the output base segment and the output size from the VM return values and the VM
 /// output builtin.
 fn get_output_info(vm: &VirtualMachine) -> Result<(usize, usize), SnOsError> {
-    let builtin_end_ptrs = vm.get_return_values(8).map_err(|e| SnOsError::CatchAll(e.to_string()))?;
+    let n_builtins = vm.get_builtin_runners().len();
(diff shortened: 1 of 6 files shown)


src/io/output.rs
```diff
@@ -2,83 +2,159 @@ use cairo_vm::types::relocatable::MaybeRelocatable;
...
+        decode_output(raw_output.into_iter())
     }
 }
 
-pub fn decode_output(mut os_output: Vec<Felt252>) -> Result<StarknetOsOutput, SnOsError> {
-    let header: Vec<Felt252> = os_output.drain(..HEADER_SIZE).collect();
+/// Gets the output base segment and the output size from the VM return values and the VM
+/// output builtin.
+fn get_output_info(vm: &VirtualMachine) -> Result<(usize, usize), SnOsError> {
+    let builtin_end_ptrs = vm.get_return_values(8).map_err(|e| SnOsError::CatchAll(e.to_string()))?;
```
notlesh, line 52: `8` should be a constant, I think it probably comes from the number of builtins, which I think you can query for

notlesh (approved):

PR #230, opened 2024-06-11
Feature: support Cairo 0 `get_tx_info()` syscall

odesenfans:
Added support for `get_tx_info()` + a test.

Note that the test needed a new function in `test_contract.cairo` as the test for this syscall was impractical for integration testing.

Issue Number: N/A

## Type

- [x] feature
- [ ] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [ ] yes
- [x] no

### DIFF
--- a/tests/integration/syscalls_tests.rs
+++ b/tests/integration/syscalls_tests.rs
@@ -1,13 +1,15 @@
 use blockifier::abi::abi_utils::selector_from_name;
 use blockifier::context::BlockContext;
 use blockifier::invoke_tx_args;
+use blockifier::test_utils::invoke::invoke_tx;
 use blockifier::test_utils::{create_calldata, NonceManager};
+use blockifier::transaction::account_transaction::AccountTransaction;
 use blockifier::transaction::test_utils;
 use blockifier::transaction::test_utils::max_fee;
 use rstest::rstest;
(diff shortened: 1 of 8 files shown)


tests/integration/syscalls_tests.rs
```diff
@@ -188,3 +190,61 @@ async fn test_syscall_get_block_timestamp_cairo0(
...
+    max_fee: Fee,
+) {
+    let initial_state = initial_state_cairo1.await;
+
+    let sender_address = initial_state.cairo1_contracts.get("account_with_dummy_validate").unwrap().address;
+    let contract_address = initial_state.cairo0_contracts.get("test_contract").unwrap().address;
+
+    let tx_version = TransactionVersion::ZERO;
+
+    let expected_chain_id = stark_felt!(1536727068981429685321u128);
```
notlesh, line 209: Is this the same as `pub const SN_GOERLI: &str = "534e5f474f45524c49";` defined in `src/config.rs`? It would be good to reuse a constant if possible or at least explain where this comes from

odesenfans: It's not the same (I think this is Sepolia) but we can use Goerli and reuse the constant :+1:

PR #235, opened 2024-06-18
Fix: load contract and class changes in OS output

odesenfans:
Problem: the `contracts` and `classes` fields of the OS output are not deserialized correctly. We need to deserialize a full structure and not just felts.

Solution: add specific deserializers for contract and class changes. Additionally, add the OS output to the return value when running the OS in order to check specific output values in changes.

Issue Number: N/A

## Type

- [ ] feature
- [x] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [x] yes, the OS output format changed.
- [ ] no

### DIFF
--- a/src/io/output.rs
+++ b/src/io/output.rs
@@ -1,3 +1,5 @@
+use std::collections::HashMap;
+
 use cairo_vm::types::relocatable::MaybeRelocatable;
 use cairo_vm::vm::runners::builtin_runner::BuiltinRunner;
 use cairo_vm::vm::vm_core::VirtualMachine;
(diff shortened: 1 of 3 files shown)


src/io/output.rs
```diff
@@ -93,18 +108,86 @@ fn get_raw_output(vm: &VirtualMachine, output_base: usize, output_size: usize) -
...
+
+fn next_as_usize<I: Iterator<Item = Felt252>>(output_iter: &mut I, item_name: &str) -> Result<usize, SnOsError> {
+    output_iter
+        .next()
+        .ok_or(SnOsError::CatchAll(format!("Could not read {item_name} segment size")))?
+        .to_usize()
+        .ok_or(SnOsError::CatchAll(format!("{item_name} segment size is too large")))
+}
+
+fn parse_contract_changes<I: Iterator<Item = Felt252>>(output_iter: &mut I) -> Result<ContractChanges, SnOsError> {
```
notlesh, line 123: Consider making this `pub`?

src/io/output.rs
```diff
@@ -93,18 +108,86 @@ fn get_raw_output(vm: &VirtualMachine, output_base: usize, output_size: usize) -
...
+    let (class_updated, nonce) = class_nonce.div_rem(&two_exp_64);
+
+    let class_hash = if class_updated != Felt252::ZERO {
+        Some(next_or_fail(output_iter, "contract change class_hash")?)
+    } else {
+        None
+    };
+
+    // unwrap() is safe because we know that n_changes fits in 64 bits by definition
+    // of the format above.
```
notlesh, line 139: This explanation could be used as the text for an `expect()` call instead of `unwrap()`, perhaps at the expense of making the binary a bit larger.

src/io/output.rs
odesenfans: Why? It's an implementation detail.

src/io/output.rs
odesenfans: Good idea.

PR #239, opened 2024-06-19
Feature: support `send_message` Cairo 0 syscall

whichqua:
Issue Number: N/A

## Type

- [ ] feature
- [ ] bugfix
- [ ] build
- [ ] docs
- [x] testing

## Breaking changes?

- [ ] yes
- [x] no

### DIFF
--- a/tests/integration/deprecated_syscalls_tests.rs
+++ b/tests/integration/deprecated_syscalls_tests.rs
@@ -491,3 +491,50 @@ async fn test_syscall_get_contract_address_cairo0(
     // We can only verify that the syscall is implemented and goes through.
     check_os_output_read_only_syscall(os_output, block_context);
 }
+
+#[rstest]
+// We need to use the multi_thread runtime to use task::block_in_place for sync -> async calls.
+#[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+async fn test_syscall_emit_event_cairo0(
+    #[future] initial_state_cairo0: StarknetTestState,
+    block_context: BlockContext,
+    max_fee: Fee,
+) {
+    let initial_state = initial_state_cairo0.await;
+
+    let sender_address = initial_state.cairo0_contracts.get("account_with_dummy_validate").unwrap().address;
+    let test_contract = initial_state.cairo0_contracts.get("test_contract").unwrap();
+    let contract_address = test_contract.address;
+
+    let tx_version = TransactionVersion::ZERO;
+    let keys = vec![stark_felt!(2019_u16), stark_felt!(2020_u16)];
+    let data = vec![stark_felt!(2021_u16), stark_felt!(2022_u16), stark_felt!(2023_u16)];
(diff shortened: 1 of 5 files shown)


tests/integration/deprecated_syscalls_tests.rs
```diff
@@ -491,3 +491,45 @@ async fn test_syscall_get_contract_address_cairo0(
...
+        max_fee,
+        sender_address: sender_address,
+        calldata: create_calldata(contract_address, "send_message", entrypoint_args),
+        version: tx_version,
+        nonce: nonce_manager.next(sender_address),
+    });
+
+    let txs = vec![tx];
+
+    let (_pie, _os_output) = execute_txs_and_run_os(
```
odesenfans, line 526: Okay so here it would be nice to check if the L1 message appears in the OS output. `StarknetOsOuput` has a `messages_to_l1` field, I would expect that the message appears there. Could you check?

odesenfans (changes requested):
Improve the test.

tests/integration/deprecated_syscalls_tests.rs
```diff
@@ -491,3 +492,103 @@ async fn test_syscall_get_contract_address_cairo0(
...
+    let output: Vec<BigUint> =
+        os_output.messages_to_l1.iter().map(|msg| BigUint::from_bytes_le(&msg.to_bytes_le())).collect();
+
+    // assert the address is returned
+    assert!(&output.contains(&address.try_into().unwrap()));
+
+    // assert the payload
+    for expected_payload in [12u16, 34] {
+        assert!(&output.contains(&expected_payload.try_into().unwrap()));
+    }
```
odesenfans, line 593: As discussed, this is too simplistic and could lead to error. Now that we know that the message looks like:
1. Sender address
2. `to_address`
3. payload length
4+. payload

I want you to test that you get exactly the expected message.

tests/integration/deprecated_syscalls_tests.rs
```diff
@@ -491,3 +492,106 @@ async fn test_syscall_get_contract_address_cairo0(
...
+    max_fee: Fee,
+) {
+    let initial_state = initial_state_cairo0.await;
+
+    let sender_address = initial_state.cairo0_contracts.get("account_with_dummy_validate").unwrap().address;
+    let test_contract = initial_state.cairo0_contracts.get("test_contract").unwrap();
+    let contract_address = test_contract.address;
+
+    // test constants
+    const ADDRESS: u16 = 1298;
```
odesenfans, line 558: What's the point of using `u16`?

odesenfans (approved):

tests/integration/deprecated_syscalls_tests.rs
odesenfans: Right, I didn't realize that someone else used `u16`. Could you use `u64` though? I'd rather avoid this type of stuff ending all over the place.

PR #247, opened 2024-06-24
Internal: move Blockifier compatibility from `SharedState` to `TestSharedState`

odesenfans:
Problem: the `SharedState` struct currently has a `contract_addresses` hashset that maps all the contracts in the state. We do not want to require an exhaustive set of all contracts when using real block data from Pathfinder.

Solution: isolate the Blockifier-related code of `SharedState` (state diff application + type conversions) to a new `TestSharedState` struct dedicated to the task. `SharedState` is now a base struct that aims to be compatible with all use cases (transaction reexecution for tests, fetch the trees from an RPC for production).

Ultimately, the goal is to remove the dependency to Blockifier from the SNOS library and only use it for testing, or hide it behind a feature flag if this is somehow requested by users. 

Issue Number: N/A

## Type

- [ ] feature
- [ ] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [x] yes, several methods are dropped from the `SharedState` API
- [ ] no

### DIFF
--- a/crates/starknet-os/src/starknet/business_logic/fact_state/mod.rs
+++ b/crates/starknet-os/src/starknet/business_logic/fact_state/mod.rs
@@ -1,3 +1,4 @@
 pub mod contract_class_objects;
 pub mod contract_state_objects;
 pub mod state;
+pub mod test_shared_state;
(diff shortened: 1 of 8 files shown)


odesenfans:
Marking this PR as obsolete, we ended up not needing this.

PR #251, opened 2024-06-28
Feature: Support `deploy_account` v1 txns

whichqua:
Problem: deploy account checks and tests are not implemented.

Solution: Implement the missing parts.

Issue Number: N/A

## Type

- [ ] feature
- [ ] bugfix
- [ ] build
- [ ] docs
- [x] testing

## Breaking changes?

- [ ] yes
- [x] no

### DIFF
--- a/src/execution/helper.rs
+++ b/src/execution/helper.rs
@@ -234,7 +234,32 @@ trait GenCallIter {
 impl GenCallIter for TransactionExecutionInfo {
     fn gen_call_iterator(&self) -> IntoIter<CallInfo> {
         let mut call_infos = vec![];
-        for call_info in self.non_optional_call_infos() {
+
+        // Determine if we are treating a DEPLOY_ACCOUNT tx. For deployments we need
+        // to order call infos differently, __validate_deploy__ is called after the constructor.
+        // See https://docs.starknet.io/architecture-and-concepts/accounts/account-functions/#overview
+        // for more details.
+        let is_deploy = match &self.execute_call_info {
+            Some(call_info) => matches!(call_info.call.entry_point_type, EntryPointType::Constructor),
+            None => false,
+        };
+
+        let call_info_iter = match is_deploy {
+            // For DEPLOY_ACCOUNT, validation is performed after executing the constructor
+            true => self
+                .execute_call_info
+                .iter()
+                .chain(self.validate_call_info.iter())
+                .chain(self.fee_transfer_call_info.iter()),
+            // For other tx types, validation comes before the execution of the call
+            false => self
+                .validate_call_info
+                .iter()
+                .chain(self.execute_call_info.iter())
+                .chain(self.fee_transfer_call_info.iter()),
+        };
+
+        for call_info in call_info_iter {
             call_infos.extend(call_info.clone().gen_call_topology());
         }
         call_infos.into_iter()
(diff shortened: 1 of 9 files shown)


src/execution/helper.rs
```diff
@@ -234,7 +234,32 @@ trait GenCallIter {
...
+        // Determine if we are treating a DEPLOY_ACCOUNT tx. For deployments we need
+        // to order call infos differently, __validate_deploy__ is called after the constructor.
+        // See https://docs.starknet.io/architecture-and-concepts/accounts/account-functions/#overview
+        // for more details.
+        let is_deploy = match &self.execute_call_info {
+            Some(call_info) => matches!(call_info.call.entry_point_type, EntryPointType::Constructor),
+            None => false,
+        };
+
+        let call_info_iter = match is_deploy {
```
notlesh, line 247: `match` on boolean is a bit unusual, you can also write as:

```rust
let foo = if condition { val1 } else { val2 }
```

This is purely cosmetic and not a big deal though :)

tests/integration/common/state.rs
```diff
@@ -295,26 +304,35 @@ impl<'a> StarknetStateBuilder<'a> {
...
-        let mut addresses: HashSet<ContractAddress> = Default::default();
+    fn fund_accounts(
+        fee_config: &FeeConfig,
+        dict_state_reader: &mut DictStateReader,
+        mut funds_per_address: HashMap<ContractAddress, Balance>,
+    ) {
         for address in dict_state_reader.address_to_class_hash.keys().chain(dict_state_reader.address_to_nonce.keys()) {
-            addresses.insert(*address);
+            if !funds_per_address.contains_key(address) {
+                funds_per_address.insert(address.clone(), fee_config.default_balance.clone());
```
notlesh, line 314: I think this would silently drop duplicates, which isn't much of a concern except that would be a real PITA to debug. A simple `log::warn` or `assert` could be appropriate

tests/integration/deploy_txn_tests.rs
```diff
@@ -0,0 +1,110 @@
...
+    block_context: BlockContext,
+    #[from(init_logging)] _logging: (),
+) -> (StarknetTestState, DeployArgs) {
+    let account_with_dummy_validate = load_cairo0_contract("account_with_dummy_validate");
+    let account_with_long_validate = load_cairo0_contract("account_with_long_validate");
+
+    // This is the hardcoded class hash of `account_with_long_validate`.
+    // Recomputing it automatically requires a significant amount of code reorganization so
+    // we hardcode it for simplicity.
+    let class_hash = class_hash!("0x067605bc345e925118dd60e09888a600e338047aa61e66361d48604ea670b709");
```
notlesh, line 41: Hmmm, could you call `to_internal_tx()` (or `to_internal_deploy_v1()`) fisrt to convert the tx and then easily call `tx_hash_deploy_account_v1()`?

notlesh: Related: https://xkcd.com/1205/

src/execution/helper.rs
odesenfans: Good point!

tests/integration/deploy_txn_tests.rs
odesenfans: It's the class hash, not the tx hash so no can do. The only method we have for that is to use the Pathfinder hash functions, and I don't want to bother with the type conversion here (or set up a storage + FFC and write a compiled class fact to it).

PR #252, opened 2024-06-28
Feature: Deploy Account v3 Tests

whichqua:
Problem: Deploy for v3 is not tested.

Solution: Implement tests for deploy v3

Issue Number: N/A

## Type

- [ ] feature
- [ ] bugfix
- [ ] build
- [ ] docs
- [x] testing

## Breaking changes?

- [ ] yes
- [x] no

### DIFF
--- a/src/hints/execution.rs
+++ b/src/hints/execution.rs
@@ -110,17 +110,15 @@ pub fn prepare_constructor_execution(
 ) -> Result<(), HintError> {
     let tx = exec_scopes.get::<InternalTransaction>(vars::scopes::TX)?;
     insert_value_from_var_name(
-        "contract_address_salt",
+        vars::ids::CONTRACT_ADDRESS_SALT,
         tx.contract_address_salt.expect("`contract_address_salt` must be present"),
         vm,
         ids_data,
         ap_tracking,
     )?;
     insert_value_from_var_name(
-        "class_hash",
-        // using `contract_hash` instead of `class_hash` as the that's how the
-        // input.json is structured
-        tx.contract_hash.expect("`contract_hash` must be present"),
+        vars::ids::CLASS_HASH,
+        tx.class_hash.expect("`class_hash` must be present"),
         vm,
         ids_data,
         ap_tracking,
--- a/src/hints/execution.rs
+++ b/src/hints/execution.rs
@@ -759,7 +759,7 @@ pub fn tx_account_deployment_data_len(
     let len = if tx.version.unwrap_or(Felt252::ZERO) < Felt252::THREE {
         0usize
     } else {
-        tx.account_deployment_data.unwrap_or_default().len()
+        tx.account_deployment_data.unwrap().len()
     };
 
     insert_value_into_ap(vm, Felt252::from(len))
@@ -781,7 +781,7 @@ pub fn tx_account_deployment_data(
         MaybeRelocatable::Int(Felt252::ZERO)
     } else {
         let data: Vec<MaybeRelocatable> =
-            tx.account_deployment_data.unwrap_or_default().iter().map(|f| MaybeRelocatable::Int(*f)).collect();
+            tx.account_deployment_data.unwrap().iter().map(|f| MaybeRelocatable::Int(*f)).collect();
         vm.gen_arg(&data)?
     };
 
--- a/src/hints/execution.rs
+++ b/src/hints/execution.rs
@@ -759,7 +759,7 @@ pub fn tx_account_deployment_data_len(
     let len = if tx.version.unwrap_or(Felt252::ZERO) < Felt252::THREE {
         0usize
     } else {
-        tx.account_deployment_data.unwrap().len()
+        tx.account_deployment_data.unwrap_or_default().len()
     };
 
     insert_value_into_ap(vm, Felt252::from(len))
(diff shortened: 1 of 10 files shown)


src/hints/execution.rs
```diff
@@ -759,7 +759,7 @@ pub fn tx_account_deployment_data_len(
     let len = if tx.version.unwrap_or(Felt252::ZERO) < Felt252::THREE {
         0usize
     } else {
-        tx.account_deployment_data.unwrap().len()
+        tx.account_deployment_data.unwrap_or_default().len()
```
odesenfans, line 762: Revert, we should fail if `account_deployment_data` is not specified here. This should actually return a `HintError::custom("tx.account_deployment_data must be specified".to_string().into_boxed_string())` instead of panicking, btw.

src/hints/execution.rs
```diff
@@ -781,7 +781,7 @@ pub fn tx_account_deployment_data(
         MaybeRelocatable::Int(Felt252::ZERO)
     } else {
         let data: Vec<MaybeRelocatable> =
-            tx.account_deployment_data.unwrap().iter().map(|f| MaybeRelocatable::Int(*f)).collect();
+            tx.account_deployment_data.unwrap_or_default().iter().map(|f| MaybeRelocatable::Int(*f)).collect();
```
odesenfans, line 784: Same, revert and return a `HintError` instead of unwrapping.

tests/integration/common/transaction_utils.rs
```diff
@@ -614,6 +647,65 @@ pub fn to_internal_invoke_v3_tx(tx: &InvokeTransactionV3) -> InternalTransaction
...
+    let nonce = felt_api2vm(tx.nonce.0);
+    let tip = felt_api2vm(tx.tip.0.into());
+
+    let nonce_data_availability_mode = Felt252::from(tx.nonce_data_availability_mode as u64);
+    let fee_data_availability_mode = Felt252::from(tx.fee_data_availability_mode as u64);
+    let resource_bounds = &tx.resource_bounds;
+
+    let paymaster_data: Vec<Felt252> = tx.paymaster_data.0.iter().map(|x| to_felt252(x.into())).collect();
+
+    let sender_address = calculate_contract_address(
```
odesenfans, line 665: This is not the `sender_address`, this is the address of the contract being deployed. This should be called `contract_address`, `deployed_contract_address` or similar.

tests/integration/common/transaction_utils.rs
```diff
@@ -614,6 +647,65 @@ pub fn to_internal_invoke_v3_tx(tx: &InvokeTransactionV3) -> InternalTransaction
...
+        tx.contract_address_salt,
+        tx.class_hash,
+        &tx.constructor_calldata,
+        ContractAddress::from(0_u8),
```
odesenfans, line 669: That's the deployer address and it shouldn't be zero in all cases. This field is provided in the tx data, please support it and parametrize the test so that we test the case where the deployer is 0 or if the deployer is the sender.

tests/integration/common/transaction_utils.rs
```diff
@@ -614,6 +647,65 @@ pub fn to_internal_invoke_v3_tx(tx: &InvokeTransactionV3) -> InternalTransaction
...
+        &calldata,
+    );
+
+    return InternalTransaction {
+        hash_value,
+        version: Some(Felt252::THREE),
+        nonce: Some(nonce),
+        sender_address: Some(sender_address),
+        entry_point_selector,
+        entry_point_type: Some("EXTERNAL".to_string()),
```
odesenfans, line 694: I think this should be "CONSTRUCTOR", not 100% sure. I don't know if it even needs to be specified at all.

odesenfans (changes requested):

src/hints/execution.rs
odesenfans: In the meantime this change should be reverted though, it's incorrect to `unwrap_or_default()`.

tests/integration/common/transaction_utils.rs
odesenfans: Mmh that's a good point, it's technically not a part of the tx but a part of the contract address... I'm a bit surprised by this, let me check the Cairo code.

odesenfans: Okay, not too sure why but this seems to be a feature used only in the `deploy` syscall and not in deploy txs. Interesting, needs to be investigated but we can keep it as it is for the tx tests here.

tests/integration/common/transaction_utils.rs
```diff
@@ -614,6 +647,66 @@ pub fn to_internal_invoke_v3_tx(tx: &InvokeTransactionV3) -> InternalTransaction
...
+        contract_address_salt,
+        class_hash,
+        &calldata,
+    );
+
+    return InternalTransaction {
+        hash_value,
+        version: Some(Felt252::THREE),
+        nonce: Some(nonce),
+        sender_address: Some(contract_address),
```
odesenfans, line 692: This should still be the sender of the tx, not the deployed contract. You can set `contract_address` to `contract_address` and `sender_address` to `sender_address`.

tests/integration/common/transaction_utils.rs
```diff
@@ -614,6 +647,66 @@ pub fn to_internal_invoke_v3_tx(tx: &InvokeTransactionV3) -> InternalTransaction
...
+
+    let contract_address = calculate_contract_address(
+        tx.contract_address_salt,
+        tx.class_hash,
+        &tx.constructor_calldata,
+        ContractAddress::from(0_u8),
+    )
+    .unwrap();
+    let contract_address_salt = felt_api2vm(tx.contract_address_salt.0);
+    let contract_address = felt_api2vm(*contract_address.0);
```
odesenfans, line 673: Nitpick but it's weird that `contract_address_salt` is converted right in the middle of generating `contract_address`.

odesenfans (changes requested):
Some more changes needed.

odesenfans (approved):

PR #256, opened 2024-07-04
Fix: Remove panicking `unwrap`s in hints execution

whichqua:
Problem: Some hints would unwrap `Option`s instead of returning an error.

Solution: Replace unwraps with early error returns

Issue Number: N/A

## Type

- [ ] feature
- [ ] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [ ] yes
- [x] no

### DIFF
--- a/src/hints/execution.rs
+++ b/src/hints/execution.rs
@@ -38,6 +38,10 @@ use crate::starkware_utils::commitment_tree::base_types::DescentMap;
 use crate::starkware_utils::commitment_tree::update_tree::{DecodeNodeCase, TreeUpdate, UpdateTree};
 use crate::utils::{execute_coroutine, get_constant};
 
+fn custom_hint_error(error: &str) -> HintError {
+    HintError::CustomHint(error.to_string().into_boxed_str())
+}
+
 pub const LOAD_NEXT_TX: &str = indoc! {r#"
         tx = next(transactions)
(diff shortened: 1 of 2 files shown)


src/hints/execution.rs
```diff
@@ -579,8 +580,7 @@ pub fn tx_entry_point_selector(
     let tx = exec_scopes.get::<InternalTransaction>(vars::scopes::TX)?;
     let entry_point_selector = tx
         .entry_point_selector
-        .ok_or(HintError::CustomHint("tx.entry_point_selector is None".to_string().into_boxed_str()))
-        .unwrap_or_default();
+        .ok_or(HintError::CustomHint("tx.entry_point_selector is None".to_string().into_boxed_str()))?;
```
odesenfans, line 583: You can use `custom_hint()` here.

src/hints/execution.rs
```diff
@@ -38,6 +38,10 @@ use crate::starkware_utils::commitment_tree::base_types::DescentMap;
 use crate::starkware_utils::commitment_tree::update_tree::{DecodeNodeCase, TreeUpdate, UpdateTree};
 use crate::utils::{execute_coroutine, get_constant};
 
+fn custom_hint_error(error: &str) -> HintError {
+    HintError::CustomHint(error.to_string().into_boxed_str())
+}
```
odesenfans, line 43: This should go in the `utils` mod as a public function.

odesenfans (changes requested):
Nice, thanks! Just two minor things to adapt.

odesenfans (approved):

PR #257, opened 2024-07-05
Internal: Clean up the transaction types passed to execution

whichqua:
Problem: We need to be able to pass different types of txs as Transaction type

Solution: Pass the general type and convert

Issue Number: N/A

## Type

- [ ] feature
- [ ] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [ ] yes
- [x] no

### DIFF
--- a/tests/integration/deprecated_syscalls_tests.rs
+++ b/tests/integration/deprecated_syscalls_tests.rs
@@ -14,6 +14,7 @@ use blockifier::test_utils::{create_calldata, NonceManager};
 use blockifier::transaction::account_transaction::AccountTransaction;
 use blockifier::transaction::test_utils;
 use blockifier::transaction::test_utils::max_fee;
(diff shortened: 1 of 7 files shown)


tests/integration/deprecated_syscalls_tests.rs
```diff
@@ -106,7 +106,7 @@ async fn test_syscall_get_block_number_cairo0(
         nonce: nonce_manager.next(sender_address),
     });
 
-    let txs = vec![tx];
+    let txs = vec![tx].into_iter().map(Into::into).collect();
```
odesenfans, line 109: Your way of converting, especially for single txs, is convoluted. You could just do:

let txs = vec![Transaction::AccountTransaction(tx)];`

odesenfans (changes requested):
The way you declare txs can be improved.

odesenfans (approved):

PR #262, opened 2024-07-12
Pathfinder Integration

notlesh:
This PR introduces Pathfinder RPC for fetching data. TODO: document better :)

## Type

- [ ] feature
- [ ] bugfix
- [ ] dev (no functional changes, no API changes)
- [ ] fmt (formatting, renaming)
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [ ] yes
- [ ] no

PR #268, opened 2024-07-17
Replace storage replay mechanism

notlesh:
This PR builds on top of #262 by exploring the ability to remove the storage value replay system, which would simplify the data we need to gather before attempting to prove a block.

Issue Number: N/A

## Type

- [ ] feature
- [ ] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [ ] yes
- [ ] no


notlesh:
Wrong repo :tada: 

(since #262 is from MSL -> KSS I need to do MSL -> MSL for this to properly be based on that branch...)

PR #277, opened 2024-07-25
Feature: add keccak syscall

HermanObst:
Adds keccak Syscall and and e2e integration test that checks correct execution

## Type
- [x] feature

## Breaking changes?
- [x] no


odesenfans (changes requested):
Looking good! Let's rebase, fix the dependencies and this should be good to go :rocket:

PR #278, opened 2024-07-26
Chore: bump blockifier to 0.7.0-dev.1

odesenfans:
Issue Number: N/A

## Type

- [ ] feature
- [ ] bugfix
- [x] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [ ] yes
- [x] no

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ anyhow = "1.0.75"
 assert_matches = "1.5.0"
 base64 = "0.21.3"
 bitvec = { version = "1.0.1", features = ["serde"] }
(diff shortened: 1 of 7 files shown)


odesenfans:
I don't know whether this is pertinent yet, but did it while attempting to reproduce the multi-call invoke tx issue. Blockifier has not had a stable release since 0.5.0 in March. Since then, many dev and rc versions got released, but we do not know how stable these are. Keep this PR open until we figure out which version of Blockifier we can use safely.

odesenfans:
Apparently v0.6.0-rc.2 is the reference version for cairo-lang 0.13.1, and 0.8.0 is the target for 0.13.2. 0.7.0-dev is therefore an in-the-middle version and can be ignored.

PR #281, opened 2024-07-26
Pathfinder Integration with Blockifier Reexecution

notlesh:
This is basically the same as #262 but rebased on top of recent work. Main work is on the `prove_block` binary, which integrates `starknet_replay` to reexecute transactions through blockifier in order to come up with an ordered list of storage reads to replay.

## Type

- [ ] feature
- [ ] bugfix
- [ ] dev (no functional changes, no API changes)
- [ ] fmt (formatting, renaming)
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [ ] yes
- [ ] no

PR #287, opened 2024-07-30
Pathfinder integration with manual block reexecution

notlesh:
This PR is the latest in an attempt to integrate Pathfinder in order to prove blocks.

Unlike #281 it does not attempt to use `starknet-replay` to do the "heavy lifting" of blockifier reexecution, and instead does this manually, meaning:

* Transactions are manually prepared for blockifier
* `StateReader` is implemented using Pathfinder's `JsonRpcClient` (using `DictReader` as an initial state / cache)

TODO:
 - [x] Remove use of `starknet-replay` (mostly `replay.rs`)
 - [x] Optimize `StateReader` impl
 - [x] Cleanly separate initial state for blockifier vs. initial state for SNOS
 - [x] Debug :sweat_smile: 
 - [x] Remove `DictStateReader` in `StateReader` impl, it's redundant with what Blockifier does internally (right?)

## Type

- [x] feature
- [ ] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [ ] yes
- [x] no


crates/bin/prove_block/src/main.rs
```diff
@@ -0,0 +1,741 @@
...
+    value: Felt252,
+}
+
+#[derive(Clone, Deserialize)]
+enum TrieNode {
+    #[serde(rename = "binary")]
+    Binary { left: Felt252, right: Felt252 },
+    #[serde(rename = "edge")]
+    Edge { child: Felt252, path: EdgePath },
+}
```
odesenfans, line 108: I think it's time to split this file into several modules. All the Pathfinder-specific types should go in a dedicated module for example.

crates/bin/prove_block/src/main.rs
```diff
@@ -0,0 +1,741 @@
...
+        address_to_class_hash.clone(),
+        address_to_nonce,
+        class_hash_to_compiled_class_hash.clone(),
+        storage_updates,
+    )
+    .await?;
+
+    // fill in class hashes for each accessed contract
+    for address in &accessed_contracts {
+        if *address != Felt252::ONE && !address_to_class_hash.contains_key(address) {
```
odesenfans, line 497: Shouldn't we do the same for address 0? Would be good to encapsulate the condition in a function to make this a bit clearer.

crates/bin/prove_block/src/main.rs
```diff
@@ -0,0 +1,741 @@
...
+        }
+        Err(_) => {
+            println!("exception:\n{:#?}", result);
+        }
+        _ => {}
+    }
+
+    result.unwrap();
+
+    Ok(())
```
odesenfans, line 740: This monster of a function needs to be split into several smaller functions to be reviewable.

crates/starknet-os/src/utils.rs
```diff
@@ -129,7 +129,7 @@ fn get_tokio_runtime_handle() -> Result<tokio::runtime::Handle, HintError> {
 
 /// Executes a coroutine from a synchronous context.
 /// Fails if no Tokio runtime is present.
-pub(crate) fn execute_coroutine<F, T>(coroutine: F) -> Result<T, HintError>
+pub fn execute_coroutine<F, T>(coroutine: F) -> Result<T, HintError>
```
odesenfans, line 132: I've been thinking that we need a `utils` crate in the workspace for this type of things, this should not be a public function of the OS library itself. But it could be a public function of a `starknet-os-utils` crate in the workspace.

odesenfans (changes requested):
Pretty hard to review this in the current state, could you clean it up a bit more?

odesenfans (approved):
LGTM.

PR #289, opened 2024-07-31
Feature: add keccak syscall

HermanObst:
Adds keccak Syscall and and e2e integration test that checks correct execution

## Type
- [x] feature

## Breaking changes?
- [x] no


crates/starknet-os/src/execution/constants.rs
```diff
@@ -68,3 +68,5 @@ pub const SEND_MESSAGE_TO_L1_GAS_COST: u64 = SYSCALL_BASE_GAS_COST + 50 * STEP_G
 pub const STORAGE_READ_GAS_COST: u64 = SYSCALL_BASE_GAS_COST + 50 * STEP_GAS_COST;
 #[allow(unused)]
 pub const STORAGE_WRITE_GAS_COST: u64 = SYSCALL_BASE_GAS_COST + 50 * STEP_GAS_COST;
+pub const KECCAK_FULL_RATE_IN_U64S: u64 = 17;
+pub const INVALID_INPUT_LENGTH_ERROR: &str = "0x000000000000000000000000496e76616c696420696e707574206c656e677468";
```
notlesh, line 72: A comment about this magic number might be nice (I'm guessing it's ASCII encoded as hex?)

crates/starknet-os/src/execution/syscall_handler.rs
```diff
@@ -529,44 +533,80 @@ impl SyscallHandler for StorageWriteHandler {
...
+        let input_len = (request.input_end - request.input_start)?;
+        // The to_usize unwrap will not fail as the constant value is 17
+        let (n_rounds, remainder) = num_integer::div_rem(input_len, KECCAK_FULL_RATE_IN_U64S.to_usize().unwrap());
+
+        if remainder != 0 {
+            return Err(SyscallExecutionError::SyscallError {
+                error_data: vec![Felt252::from_hex_unchecked(INVALID_INPUT_LENGTH_ERROR)],
+            });
+        }
+        let gas_cost = n_rounds.to_u64().unwrap() * KECCAK_ROUND_COST_GAS_COST;
```
notlesh, line 576: It's not clear if this `unwrap()` could fail.

If `request.input_end` and `request.input_start` are completely arbitrary, then `input_len / 17` could be `> u64::MAX`.

If that's possible, this should be handled more gracefully, and if it's not possible a comment about why would help.

crates/starknet-os/src/execution/syscall_handler.rs
```diff
@@ -529,44 +533,80 @@ impl SyscallHandler for StorageWriteHandler {
...
+        if gas_cost > *remaining_gas {
+            return Err(SyscallExecutionError::OutOfGas { remaining_gas: (*remaining_gas) });
+        }
+        *remaining_gas -= gas_cost;
```
notlesh, line 580: This should be checked math to ensure that it doesn't underflow, although I suspect that an underflow would cause some other failure...

Cargo.toml
```diff
@@ -21,7 +21,7 @@ assert_matches = "1.5.0"
 base64 = "0.21.3"
 bitvec = { version = "1.0.1", features = ["serde"] }
 # Point to the latest commit of branch msl/snos-0.6.0-rc.2
-blockifier = { git = "https://github.com/Moonsong-Labs/blockifier", rev = "30bb3d29b695ec4e0e9e6dccb62737f8f9474a54", features = ["testing"] }
+blockifier = { git = "https://github.com/Moonsong-Labs/blockifier", rev = "98d764a270c6a1b280edb94c973171027e8611fc", features = ["testing"] }
```
odesenfans, line 24: Where did you find this revision? Latest on `msl/snos-0.6.0-rc.2` is [983e3b1cdb6621c5e6fa600f47cd69bf9e287621](https://github.com/Moonsong-Labs/blockifier/commits/msl/snos-0.6.0-rc.2/).

crates/starknet-os/src/execution/constants.rs
HermanObst: @notlesh good suggestion! Just added it!

crates/starknet-os/src/execution/syscall_handler.rs
odesenfans, line 576: Why do you compute this on `u64`? Not that we really risk an overflow, but out of precaution there is nothing wrong doing that on felts/biguints.

crates/starknet-os/src/execution/syscall_handler.rs
HermanObst: @notlesh how this can underflow if im checking `if gas_cost > *remaining_gas`?

notlesh (changes requested):
Looks good, but there are a few arithmetic details that need careful consideration in the syscall

crates/starknet-os/src/execution/syscall_handler.rs
notlesh: Oh, you're right. Sorry, I've been conditioned to avoid the arithmetic operators altogether in favor of something like this: https://docs.rs/num/latest/num/trait.CheckedSub.html

Your check is sufficient :+1:

crates/starknet-os/src/execution/syscall_handler.rs
HermanObst: You are right! Done :)

HermanObst: @odesenfans I receive `remaining_gas` as a `u64`.  Thats why im casting `gas_cost`

notlesh (approved):

Cargo.toml
HermanObst: You are right! I put the hash commit of my PR, but forgot to update it after the squash.

https://github.com/Moonsong-Labs/blockifier/pull/4/commits/98d764a270c6a1b280edb94c973171027e8611fc

Fixed!

PR #290, opened 2024-08-01
Feature: generic CASM class

odesenfans:
Problem: we use different CASM contract classes depending on the use
case. We serialize/deserialize with the `cairo-lang` implementation, we
reexecute txs with the Blockifier implementation and we compute class
hashes through Pathfinder. Currently, we convert explicitly between
these types/crates. This makes the code harder to read and maintain.

Solution: introduce a generic class that is able to convert to/from all
the types we use under the hood. This puts a lid on the problem and
reduces the amount of code to change if things improve in the ecosystem.

Issue Number: N/A

## Type

- [x] feature
- [ ] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [x] yes, the new type appears in `StarknetOsInput`.
- [ ] no

### DIFF
--- a/crates/starknet-os-types/src/contract_class.rs
+++ b/crates/starknet-os-types/src/contract_class.rs
@@ -1,0 +1,220 @@
+use std::cell::OnceCell;
+use std::rc::Rc;
+
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+
+use crate::hash::GenericClassHash;
+
+pub type CairoLangCasmClass = cairo_lang_starknet_classes::casm_contract_class::CasmContractClass;
+pub type BlockifierCasmClass = blockifier::execution::contract_class::ContractClassV1;
+
+#[derive(thiserror::Error, Debug)]
+pub enum ContractClassError {
+    #[error("Internal error: no type conversion is possible to generate the desired effect.")]
+    NoPossibleConversion,
+
+    #[error("Could not build Blockifier contract class")]
+    BlockifierConversionError,
+
+    #[error(transparent)]
+    SerdeError(#[from] serde_json::Error),
+}
+
+/// A generic contract class that supports conversion to/from the most commonly used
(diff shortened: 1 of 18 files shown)


crates/starknet-os-types/src/contract_class.rs
```diff
@@ -0,0 +1,170 @@
...
+    SerdeError(#[from] serde_json::Error),
+}
+
+/// A generic contract class that supports conversion to/from the most commonly used
+/// contract class types in Starknet and provides utility methods.
+/// Operations are implemented as lazily as possible, i.e. we only convert
+/// between different types if strictly necessary.
+/// Fields are boxed in an RC for cheap cloning.
+#[derive(Debug, Clone)]
+pub struct GenericCasmContractClass {
```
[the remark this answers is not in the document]
odesenfans: No because I don't want to instantiate all the types inside this in all code paths. There may be cases where we can just instantiate the `cairo-lang` variant and do everything with it, avoiding the generation of the Blockifier variant.

odesenfans:
Depends on #292.

crates/starknet-os-types/src/contract_class.rs
```diff
@@ -0,0 +1,218 @@
...
+    blockifier_contract_class: OnceCell<Rc<BlockifierCasmClass>>,
+    cairo_lang_contract_class: OnceCell<Rc<CairoLangCasmClass>>,
+    serialized_class: OnceCell<Vec<u8>>,
+    class_hash: OnceCell<GenericClassHash>,
```
notlesh, line 33: This type is small, is it worth the overhead of `OnceCell`?

notlesh (approved):

crates/starknet-os-types/src/contract_class.rs
odesenfans: The memory footprint is small, the computation to get the value is not.

PR #292, opened 2024-08-01
Internal: introduce the `starknet-os-types` crate

odesenfans:
Problem: we need a single crate that defines useful types for all parts of the project. Useful types should be available from all the other crates.

Solution: introduce the `starknet-os-types` crate. Moved the `Hash` type there as a demonstration and preparation for upcoming PRs.

Issue Number: N/A

## Type

- [ ] feature
- [ ] bugfix
- [x] dev (no functional changes, no API changes)
- [ ] fmt (formatting, renaming)
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [ ] yes
- [x] no

PR #293, opened 2024-08-01
Internal: derive `Copy` for `Hash`

odesenfans:
Problem: as all the 32-byte wide types we use implement Copy, there is no
reason for `Hash` not to.

Issue Number: N/A

## Type

- [ ] feature
- [ ] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [ ] yes
- [x] no


odesenfans:
Depends on #292.

PR #295, opened 2024-08-05
Feature: generic deprecated compiled class

odesenfans:
Problem: as for CASM classes, we use many different deprecated compiled
class types, creating a mess.

Solution: introduce `GenericDeprecatedCompiledClass`, a type that is
able to convert to all useful deprecated compiled class types and
to compute class hashes.

Issue Number: N/A

## Type

- [x] feature
- [ ] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [ ] yes
- [x] no


odesenfans:
Depends on #290.

PR #300, opened 2024-08-06
Feature: RPC-based state reader for Blockifier reexecution

odesenfans:
Problem: we need a way to reexecute blocks with Blockifier.

Solution: implement the `StateReader` trait using the `starknet-rs` crate to fetch all the relevant data from a Pathfinder node.

Issue Number: N/A

## Type

- [x] feature
- [ ] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [ ] yes
- [x] no


crates/starknet-replay/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "starknet-replay"
```
notlesh, line 2: there are already 2 existing crates/projects called `starknet-replay`, I'd suggest not creating a 3rd ;)

odesenfans: I need a better name :(

crates/starknet-replay/src/rpc_state_reader.rs
```diff
@@ -0,0 +1,144 @@
...
+use starknet_api::core::{ClassHash, CompiledClassHash, ContractAddress, Nonce};
+use starknet_api::hash::StarkFelt;
+use starknet_api::state::StorageKey;
+use starknet_os_types::deprecated_compiled_class::GenericDeprecatedCompiledClass;
+use starknet_os_types::hash::GenericClassHash;
+use starknet_os_types::sierra_contract_class::GenericSierraContractClass;
+
+use crate::utils::{execute_coroutine, felt_api2vm, felt_vm2api};
+
+pub struct AsyncRpcStateReader<T>
```
notlesh, line 16: Is `T` here just for the `JsonRpcClient<T>`? We could give it the concrete type here of `HttpTransport` to make our lives easier unless we have a good reason to keep that generic.

notlesh: Meh, this wasn't a big deal, I wouldn't change it unless it makes sense

crates/starknet-replay/src/rpc_state_reader.rs
```diff
@@ -0,0 +1,144 @@
...
+
+fn to_state_err<E: ToString>(e: E) -> StateError {
+    StateError::StateReadError(e.to_string())
+}
+
+impl<T> AsyncRpcStateReader<T>
+where
+    T: JsonRpcTransport + Sync + Send + 'static,
+{
+    async fn get_storage_at_async(&self, contract_address: ContractAddress, key: StorageKey) -> StateResult<StarkFelt> {
```
notlesh, line 45: I would make these `pub` if the `Async` in `AsyncRpcStateReader` is to be publicly meaningful :)

crates/starknet-replay/Cargo.toml
notlesh: maybe `rpc-replay`? not great, but at least it's unique

odesenfans: Sold!

crates/starknet-replay/src/rpc_state_reader.rs
odesenfans: Makes sense!

notlesh (approved):

PR #303, opened 2024-08-07
Feature: support visited PCs

odesenfans:
Problem: the code needed to only partially load programs based on visited PCs was left as a TODO. We were wondering if this code could be the cause of some hashing issues.

Solution: implement bytecode segment loading properly. We should now support visited PCs to a reasonable extent.

Issue Number: N/A

## Type

- [x] feature
- [ ] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [ ] yes
- [x] no


tests/integration/common/block_utils.rs
```diff
@@ -133,11 +133,16 @@ where
     };
 
     let mut ffc = blockifier_state.state.ffc.clone();
+
+    // Blockifier provides a class hash -> visited PCs map, but we need
+    // the compiled class hash -> visited PCs map.
```
notlesh: This makes sense -- the OS needs to know the bytecode it must load for each compiled class.

Question, though: for each compiled contract, don't we want the union of individual sets of visited PCs?

In other words, if two txns invoke the same compiled contract A with visited PCs B and C, we want the resulting HashMap to contain `A = union(B, C)`.

I think this code will **overwrite** entries when multiple `class_hash`es map to a single `compiled_class_hash`.

odesenfans: I don't think it's possible for two different Sierra classes to generate the same CASM, but maybe I'm missing something?

notlesh: Interesting, I'm not sure. Here's a quick look at the class_hash definition:

```
class_hash = ℎ(
    contract_class_version,
    external_entry_points,
    l1_handler_entry_points,
    constructor_entry_points,
    abi_hash,
    sierra_program_hash
)
```

I think I could force this sort of collision by producing a different set of entry points on the same sierra program (assuming that's a valid thing to do). In that case, we would get a different `class_hash` with the same `sierra_program_hash`.

notlesh: One simple way to ensure we catch this if it ever happens is to assert that we aren't overwriting a value. The easiest way to do this is probably to compare the sizes of the original map (key: `ClassHash`) to the mapped one (key: `CompiledClassHash`). If they aren't the same size, warn/panic/etc.

HermanObst (commented):
Looking good 🚀 
@odesenfans correct me if im wrong: 
Now adding the visited PCs to the `StarknetOsInput` in our integration test would mean that all our e2e test now are using them?

odesenfans:
@HermanObst yes, for integration tests they're added automatically now!

HermanObst:
> @HermanObst yes, for integration tests they're added automatically now!

Awesome! So before how the test were passing? 
Is it the case that the OS has some default to prove the whole contract bytecode (without the optimization of the visited entrypoint) 🤔

odesenfans:
Yes, without visited PCs we just default to providing the whole bytecode.

notlesh (approved):

PR #306, opened 2024-08-09
Reexecute block / Pathfinder Integration continuation

notlesh:
This PR continues improving the `prove_block` binary. Specifically:

* More data fetched through RPC
* We now pass all OS txn executions on the block we've been focusing on (`Sepolia` block `76793`)
* Class hashing and compiled class hashing are consistent with OS
    * Including the `starknet-core` types used by Pathfinder
* Class bytecode is properly loaded, including for contracts enountered during subcalls
    * (deprecated contracts still TODO)
* Visited-PCs (and bytecode segmentation) supported

Issue Number: N/A

## Type

- [x] feature
- [x] bugfix
- [ ] build
- [ ] docs
- [x] testing

## Breaking changes?

- [ ] yes
- [x] no


crates/bin/prove_block/src/main.rs
```diff
@@ -189,20 +193,62 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
+                    ExecuteInvocation::Reverted(_) => unimplemented!("handle reverted invoke trace"),
+                }
+                if let Some(inv) = invoke_trace.fee_transfer_invocation {
+                    process_function_invocations(inv, &mut contracts_subcalled, 0);
+                }
             }
-        };
+            _ => unimplemented!("process other txn traces"),
+        }
+    }
```
HermanObst, line 232: This will fails as soon as we try with other block. But ok for now :) 
Already created a ticket for Fernando :P

crates/bin/prove_block/src/main.rs
```diff
@@ -189,20 +193,62 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
+    log::info!("Contract subcalls:");
+    for addr in &contracts_subcalled {
+        log::info!("    {:x}", addr);
+    }
```
HermanObst, line 236: Remove?

crates/bin/prove_block/src/main.rs
```diff
@@ -290,8 +340,24 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
+    for transaction in transactions.iter_mut() {
+        if let Some(sender_address) = transaction.sender_address {
+            let class_hash = address_to_class_hash
+                .get(&sender_address)
+                .expect("should have a class_hash for each known contract addresses at this point");
+            log::info!("Filling in class_hash {:x} for txn", class_hash);
+            transaction.class_hash = Some(*class_hash);
+        } else {
+            // TODO: are there txn types which wouldn't have a sender address?
+            log::warn!("Found transaction without sender_address");
```
HermanObst, line 355: Did you encounter some of these?
maybe its better to just put an unimplemented?

crates/bin/prove_block/src/main.rs
```diff
@@ -290,8 +340,24 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
+        }
+    }
+
     // write facts from proof
-    for proof in storage_proofs.values() {
+    for proof in storage_proofs.values().chain(previous_storage_proofs.values()) {
```
HermanObst, line 360: Why do we need this?

crates/bin/prove_block/src/main.rs
```diff
@@ -321,146 +387,113 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
+        let previous_storage_proof =
+            previous_storage_proofs.get(&contract_address).expect("failed to find previous storage proof");
 
-        updates.push((contract_address_biguint, contract_state.clone()));
-        contract_states.insert(storage_diff_item.address, contract_state);
+        let previous_tree = PatriciaTree { root: previous_storage_proof.state_commitment.into(), height: Height(251) };
 
-        let contract_storage =
-            OsSingleStarknetStorage::new(initial_tree, updated_tree, &[], initial_state.ffc.clone()).await?;
+        let contract_storage = ProverPerContractStorage::new(previous_block_id, contract_address, provider_url.clone());
```
HermanObst, line 399: Why do we need this? Can you add a comment?
@notlesh

crates/bin/prove_block/src/main.rs
```diff
@@ -321,146 +387,113 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
-        let updated_tree = contract_state.storage_commitment_tree.clone();
-
-        if storage_proof.class_commitment != contract_state.storage_commitment_tree.root.into() {
-            log::error!(
-                "expected class_commitment != computed class_commitment ({:?} != {:?})",
-                storage_proof.class_commitment,
-                contract_state.storage_commitment_tree.root.clone()
-            );
-        }
+    for (contract_address, _storage_proof) in storage_proofs {
```
HermanObst, line 393: Please add a comment on the general idea on what this for is doing

crates/bin/prove_block/src/main.rs
```diff
@@ -321,146 +387,113 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
         contract_storages.insert(contract_address, contract_storage);
-    }
 
-    // insert ContractState for any contract that received a nonce update but not a storage update
-    for nonce_update in state_update.state_diff.nonces {
-        if let std::collections::hash_map::Entry::Vacant(e) = contract_states.entry(nonce_update.contract_address) {
-            let mut contract_state = ContractState::empty(Height(251), &mut initial_state.ffc).await?;
+        log::debug!("contract address: {}", contract_address.to_hex_string());
+        let (class_hash, previous_nonce) = if [Felt252::ZERO, Felt252::ONE].contains(&contract_address) {
+            (Felt252::ZERO, Felt252::ZERO)
```
HermanObst, line 404: Which case are you handling here? Can you add a comment? :)

crates/bin/prove_block/src/main.rs
```diff
@@ -321,146 +387,113 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
     for contract_address in &accessed_contracts {
-        // TODO: dedupe with query in build_initial_state()...
         if let Ok(class_hash) = provider.get_class_hash_at(BlockId::Number(block_number), contract_address).await {
+            let contract_class = provider.get_class(BlockId::Number(block_number), class_hash).await?;
+            let generic_sierra_cc = match contract_class {
+                starknet::core::types::ContractClass::Sierra(flattened_sierra_cc) => {
+                    GenericSierraContractClass::from(flattened_sierra_cc)
+                }
+                starknet::core::types::ContractClass::Legacy(_) => {
+                    unimplemented!("Fixme: Support legacy contract class")
```
HermanObst, line 431: Already create a ticket for this

crates/bin/prove_block/src/main.rs
```diff
@@ -321,146 +387,113 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
-                }
-            } else {
-                log::warn!("No class available for contract {}", contract_address);
+            let generic_cc = generic_sierra_cc.compile()?;
+
+            log::debug!("============================");
+            log::debug!("class_hash (from RPC): {:x?}", class_hash);
+            log::debug!("class_hash (computed): {:x?}", generic_sierra_cc.class_hash()?);
+
+            // TODO: it seems that this ends up computing the wrong class hash...
```
HermanObst, line 443: remove

crates/bin/prove_block/src/main.rs
```diff
@@ -321,146 +387,113 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
+            let (contract_class_hash, compiled_contract_hash) = write_class_facts(
+                generic_sierra_cc,
+                generic_cc.clone(),
+                &mut initial_state.ffc.clone_with_different_hash::<PoseidonHash>(),
```
HermanObst, line 447: Why do we need to use a different hash function?

crates/bin/prove_block/src/main.rs
```diff
@@ -321,146 +387,113 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
+                compiled_contract_hash
+            );
+
+            // let sierra_cc_hash = generic_sierra_cc.class_hash()?;
+            let sierra_cc_hash = generic_cc.class_hash()?;
+            log::debug!("sierra_cc_hash: {:x?}", sierra_cc_hash);
+
+            let class_hash_as_snos_hash: starknet_os_types::hash::Hash = class_hash.into();
+            if contract_class_hash != class_hash_as_snos_hash {
+                log::warn!("FFC computed different hash ({:x?}) than expected ({:?})", contract_class_hash, class_hash);
```
HermanObst, line 466: What this means?

HermanObst: I would remove it. Or make it panic.

crates/bin/prove_block/src/main.rs
```diff
@@ -321,146 +387,113 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
             }
         } else {
             log::warn!("No class hash available for contract {}", contract_address);
         };
     }
 
+    // query storage proofs for each accessed contract
```
HermanObst, line 473: Its redundant, but add that you are query storage proofs for the class trie. (as difference as previous ones)

crates/bin/prove_block/src/main.rs
```diff
@@ -321,146 +387,113 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
+                TrieNode::Binary { left, right } => {
+                    // log::info!("writing binary node...");
+                    let fact = BinaryNodeFact::new((*left).into(), (*right).into())?;
+                    fact.set_fact(&mut initial_state.ffc_for_class_hash).await?;
+                }
+                TrieNode::Edge { child, path } => {
+                    // log::info!("writing edge node...");
+                    let fact = EdgeNodeFact::new((*child).into(), NodePath(path.value.to_biguint()), Length(path.len))?;
+                    fact.set_fact(&mut initial_state.ffc_for_class_hash).await?;
+                }
```
HermanObst, line 492: Here we dont need the previous ones (as with contract proofs)?

Not necessary to add now, but im curious :) 

If we need, lests make sure to add this in the ticket that refers to class data stuff

crates/bin/prove_block/src/main.rs
```diff
@@ -478,40 +511,74 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
         .map(|(class_hash, visited_pcs)| {
-            let class_hash_felt = felt_api2vm(class_hash.0);
-            let compiled_class_hash_felt = class_hash_to_compiled_class_hash.get(&class_hash_felt).unwrap();
-            (*compiled_class_hash_felt, visited_pcs.iter().copied().map(Felt252::from).collect::<Vec<_>>())
+            (felt_api2vm(class_hash.0), visited_pcs.iter().copied().map(Felt252::from).collect::<Vec<_>>())
         })
         .collect();
 
+    log::debug!("class_hash_to_compiled_class_hash ({}):", class_hash_to_compiled_class_hash.len());
+    for (ch, cch) in &class_hash_to_compiled_class_hash {
```
HermanObst, line 523: put better names to this variables

crates/bin/prove_block/src/main.rs
```diff
@@ -478,40 +511,74 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
+    //         updated_state.contract_classes.clone().expect("updated state should have class trie"),
+    //         &contract_indices,
+    //         &mut initial_state.ffc.clone_with_different_hash::<PoseidonHash>(),
+    //     )
+    //     .await
+    //     .unwrap_or_else(|e| panic!("Could not create contract class commitment info: {:?}", e));
+
     let os_input = StarknetOsInput {
-        contract_state_commitment_info,
+        contract_state_commitment_info: Default::default(),
```
HermanObst, line 570: Why are you putting a default?

HermanObst (changes requested):

crates/bin/prove_block/src/main.rs
```diff
@@ -189,20 +193,62 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
 
-    let state_update =
-        match provider.get_state_update(BlockId::Number(block_number)).await.expect("Failed to get state update") {
-            MaybePendingStateUpdate::Update(update) => update,
-            MaybePendingStateUpdate::PendingUpdate(_) => {
-                panic!("Block is still pending!")
+    let state_update = match provider.get_state_update(block_id).await.expect("Failed to get state update") {
+        MaybePendingStateUpdate::Update(update) => update,
+        MaybePendingStateUpdate::PendingUpdate(_) => {
+            panic!("Block is still pending!")
```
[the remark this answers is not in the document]
notlesh: dunno, I didn't write that ;) I'm guessing because we don't support a pending block right now (or intend to). I suspect that properly supporting this would require additional logic for retrying the block until it's no longer pending, etc. and that the caller could handle this detail more gracefully anyway.

A pending block usually means that consensus hasn't been reached for it, but I'm not sure precisely what it means in this context.

My suggestion would be to return an error more gracefully, but I'm not sure how to even produce this case anyway :shrug:

crates/bin/prove_block/src/main.rs
notlesh: By `previous` are you referring to using `block_number` instead of `block_number - 1`?

That's a good catch, and probably something really subtle since the class isn't likely to change from block to block...

crates/bin/prove_block/src/main.rs
notlesh: This isn't relevant until the end of the OS when we compute the state roots, which is coming $soon

crates/bin/prove_block/src/main.rs
notlesh: Do you mean that `storage proofs` should be `class proofs`?

crates/bin/prove_block/src/main.rs
notlesh: I imagine it's because we need to write facts for all the storage proofs we fetched. Specifically, we need to prove any reads that we did (`block_number - 1`) and any writes that we do (`block_number`).

@odesenfans wrote this, so maybe he has more context...

crates/bin/prove_block/src/main.rs
notlesh: There are magic contracts at addresses `0x0` and `0x1` and I think this is handling these by setting class hash and nonce to 0, and that the OS expects this to be the case (@odesenfans wrote this so I can't be 100% sure).

I suggest we write a utility fn to detect if a given contract address is a magical one for clarity and extensibility. But, as you suggest, this could also use a comment in any case.

crates/bin/prove_block/src/main.rs
notlesh: I'm not sure what you mean, a comment about why we need storage for each contract?

crates/bin/prove_block/src/main.rs
notlesh: Because the class trie uses Poseidon hashing whereas the storage trie uses Pedersen hashing.

But -- `SharedState` has a `ffc_for_class_hash` which would be better to use here (TODO)

crates/bin/prove_block/src/main.rs
```diff
@@ -189,20 +193,62 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
+    };
+
+    // utility to extract all contract address in a nested call structure
+    fn process_function_invocations(inv: FunctionInvocation, contracts: &mut HashSet<Felt252>, nesting_level: u64) {
+        log::info!("process_function_inv: contract: {:x} at level {}", inv.contract_address, nesting_level);
+        contracts.insert(inv.contract_address);
+        for call in inv.calls {
+            process_function_invocations(call, contracts, nesting_level + 1);
+        }
+    }
```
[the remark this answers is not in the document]
notlesh: I'm a fan of closures like this being kept close to their call site for better readability and modularity, but I couldn't actually write this as a closure and `main` is ridiculously long anyway, so I'll move it :)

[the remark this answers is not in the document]
notlesh: https://github.com/Moonsong-Labs/snos/pull/19/commits/6fa40cd7b5cd65050539c9836291c27b70a40566

crates/bin/prove_block/src/main.rs
notlesh: https://github.com/Moonsong-Labs/snos/pull/19/commits/bf9662e0121027befbe9e33c9db4d112f55c8347

crates/bin/prove_block/src/main.rs
notlesh: I didn't encounter any, but it's an `Option` so I'm guessing it's possible for some older txn versions or something like that.

notlesh: Converted to `unimplemented!()` in https://github.com/Moonsong-Labs/snos/pull/19/commits/b514f57365087bb88a617b237778c0a9d20a28f0

crates/bin/prove_block/src/main.rs
notlesh: https://github.com/Moonsong-Labs/snos/pull/19/commits/ff022701ea43134303fb2f6bc055d80396700b06

crates/bin/prove_block/src/main.rs
notlesh: Changed to `ffc_for_class_hash` in https://github.com/Moonsong-Labs/snos/pull/19/commits/8575e0822b76949eac25112046d7f061727c364d

crates/bin/prove_block/src/main.rs
notlesh: Removed this and other verbose class hash logging in https://github.com/Moonsong-Labs/snos/pull/19/commits/f032b2fc3de9b28cf6c0b690ea813e69b16d223c

crates/bin/prove_block/src/main.rs
notlesh: Changed to block-1 here: https://github.com/Moonsong-Labs/snos/pull/19/commits/3386814ebe0b2624604a8e7c8a5eb68a1164ab76

That said, I wonder if we actually need both like we do with the storage proofs... @odesenfans WDYT?

crates/bin/prove_block/src/main.rs
notlesh: Also, see all the commented-out code above this -- it'll be implemented soon

crates/bin/prove_block/src/main.rs
HermanObst: Yup! But in the blocks that decalre stuff can happen! Maybe leave a comment or TODO just for us to be aware of this

crates/bin/prove_block/src/main.rs
notlesh: https://github.com/Moonsong-Labs/snos/pull/19/commits/e1f040a66cf661659dc84d0a95ab22162255b4cd

crates/bin/prove_block/src/main.rs
HermanObst: something like: `query class trie storage proofs for each accessed contract`

crates/bin/prove_block/src/main.rs
HermanObst: @notlesh yes for now just adding the comment stating that there are some special contracts is ok :)

crates/bin/prove_block/src/main.rs
notlesh: Oh, good point, that's a simple case where this would be relevant.

notlesh: https://github.com/Moonsong-Labs/snos/pull/19/commits/7271c799dbcbd32552e631e709f9901c86c610fa

PR #313, opened 2024-08-22
Fix: decompress starknet-core legacy class

odesenfans:
Problem: the conversion from starknet-core `CompressedLegacyContractClass` in `GenericDeprecatedContractClass` is untested and broken. This class cannot be used interchangeably with other deprecated contract classes because the `program` field is deserialized as a gzip bytes vector.

Solution: add conversion functions to decompress the program field and generate a `LegacyContractClass` from it before storing it in the generic class.

Issue Number: N/A

## Type

- [ ] feature
- [x] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [ ] yes
- [x] no

### DIFF
--- a/crates/starknet-os-types/src/starknet_core_addons.rs
+++ b/crates/starknet-os-types/src/starknet_core_addons.rs
@@ -91,6 +91,10 @@ pub enum LegacyContractDecompressionError {
     Decompression(#[from] std::io::Error),
 }
 
+/// Decompresses a compressed legacy contract class.
+/// Compressed classes store the `program` field as a gzipped vector. This function
+/// decompresses the field and performs additional type conversions for the `abi` and
+/// `entry_points_by_type` fields.
 pub(crate) fn decompress_starknet_core_contract_class(
     compressed_legacy_class: CompressedLegacyContractClass,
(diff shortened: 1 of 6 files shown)


HermanObst (approved):
LGTM!

crates/starknet-os-types/src/starknet_core_addons.rs
```diff
@@ -0,0 +1,141 @@
...
+    Serde(#[from] serde_json::Error),
+
+    #[error("Failed to decompress the program: {0}")]
+    Decompression(#[from] std::io::Error),
+}
+
+pub(crate) fn decompress_starknet_core_contract_class(
+    compressed_legacy_class: CompressedLegacyContractClass,
+) -> Result<LegacyContractClass, LegacyContractDecompressionError> {
+    // Compressed classes have the program as a gzipped vector.
```
notlesh, line 97: This description of the problem is good but is hard to find. It seems that the `TryFrom` impl is the main entry for using this decompression, which isn't very self-descriptive.

odesenfans: I improved the doc a little.

PR #321, opened 2024-08-22
Chore: refactor prove block state functions

HermanObst:
Problem: we need to modularize prove_block binary

Solution: modularize

⚠️ Please run prove_block with our successful block to ensure that we dont get regressions

## Breaking changes?

- [ ] yes
- [X] no

### DIFF
--- a/crates/bin/prove_block/src/state_utils.rs
+++ b/crates/bin/prove_block/src/state_utils.rs
@@ -4,10 +4,11 @@ use std::hash::Hash;
 
 use cairo_vm::Felt252;
 use pathfinder_crypto::Felt;
-use starknet::core::types::{BlockId, MaybePendingStateUpdate};
+use starknet::core::types::{BlockId, MaybePendingStateUpdate, StateDiff};
 use starknet::providers::jsonrpc::HttpTransport;
 use starknet::providers::{JsonRpcClient, Provider};
 use starknet_os::crypto::pedersen::PedersenHash;
+use starknet_os::io::InternalTransaction;
 use starknet_os::starknet::business_logic::fact_state::contract_class_objects::{
     get_ffc_for_contract_class_facts, ContractClassLeaf,
 };
@@ -19,13 +20,14 @@ use starknet_os::starkware_utils::commitment_tree::binary_fact_tree::BinaryFactT
 use starknet_os::storage::storage::FactFetchingContext;
 
 use crate::rpc_utils::{CachedRpcStorage, RpcStorage};
+use crate::utils::get_subcalled_contracts_from_tx_traces;
 
 struct ProcessedStateUpdate {
     // TODO: Use more descriptive types
-    address_to_class_hash: HashMap<Felt, Felt>,
-    address_to_nonce: HashMap<Felt, Felt>,
-    class_hash_to_compiled_class_hash: HashMap<Felt, Felt>,
-    storage_updates: HashMap<Felt, HashMap<Felt, Felt>>
+    address_to_class_hash: HashMap<Felt252, Felt252>,
+    address_to_nonce: HashMap<Felt252, Felt252>,
+    class_hash_to_compiled_class_hash: HashMap<Felt252, Felt252>,
+    storage_updates: HashMap<Felt252, HashMap<Felt252, Felt252>>
 }
 
 // build state representing the end of the previous block on which the current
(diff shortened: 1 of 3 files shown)


crates/bin/prove_block/src/state_utils.rs
```diff
@@ -14,9 +16,21 @@ use starknet_os::starknet::business_logic::fact_state::state::SharedState;
...
+
+#[derive(Clone)]
+pub struct ProcessedStateUpdate {
+    // TODO: Use more descriptive types
+    pub address_to_class_hash: HashMap<Felt252, Felt252>,
+    pub address_to_nonce: HashMap<Felt252, Felt252>,
+    pub class_hash_to_compiled_class_hash: HashMap<Felt252, Felt252>,
+    pub storage_updates: HashMap<Felt252, HashMap<Felt252, Felt252>>,
+    pub accessed_addresses: HashSet<Felt252>,
+}
```
odesenfans, line 33: Unrelated to your PR, but yeah we need to use better typing here. This should be doable once I get my dependency PR to work.

crates/bin/prove_block/src/state_utils.rs
```diff
@@ -14,9 +16,21 @@ use starknet_os::starknet::business_logic::fact_state::state::SharedState;
...
 use starknet_os::starkware_utils::commitment_tree::binary_fact_tree::BinaryFactTree;
+use starknet_os::starkware_utils::commitment_tree::patricia_tree::patricia_tree::PatriciaTree;
 use starknet_os::storage::storage::FactFetchingContext;
 
-use crate::rpc_utils::CachedRpcStorage;
+use crate::rpc_utils::{CachedRpcStorage, RpcStorage};
+use crate::utils::get_subcalled_contracts_from_tx_traces;
+
+#[derive(Clone)]
+pub struct ProcessedStateUpdate {
```
odesenfans, line 26: What is a "processed" state update?

crates/bin/prove_block/src/state_utils.rs
```diff
@@ -26,81 +40,205 @@ use crate::rpc_utils::CachedRpcStorage;
...
-            contract_classes: updated_contract_classes,
+            contract_classes: updated_global_class_root,
             ffc,
             ffc_for_class_hash: ffc_for_contract_class,
             contract_addresses: Default::default(),
         },
     ))
 }
+
+fn update_empty_class_trie_with_block_incoming_changes(
```
odesenfans, line 107: Document every new function.

crates/bin/prove_block/src/state_utils.rs
```diff
@@ -26,81 +40,205 @@ use crate::rpc_utils::CachedRpcStorage;
...
+    processed_state_update: &ProcessedStateUpdate,
+) -> Vec<(BigUint, ContractClassLeaf)> {
+    let classes_modifications = processed_state_update
```
odesenfans, line 110:
```suggestion
    let class_modifications = processed_state_update
```

Nitpick: we always use `class`, singular.

crates/bin/prove_block/src/state_utils.rs
odesenfans: Looking at the implementation the name of the function does not reflect what it does. This could be named `get_class_modifications`, `format_class_modifications`, `get_class_updates`, ...

crates/bin/prove_block/src/state_utils.rs
```diff
@@ -26,81 +40,205 @@ use crate::rpc_utils::CachedRpcStorage;
...
+    classes_modifications
+}
+
+async fn update_empty_contract_state_with_block_incoming_changes(
+    empty_contract_states: &mut HashMap<TreeIndex, ContractState>,
+    processed_state_update: &ProcessedStateUpdate,
+    provider: &JsonRpcClient<HttpTransport>,
+    block_id: BlockId,
+    ffc: &mut FactFetchingContext<CachedRpcStorage, PedersenHash>,
+) -> Result<HashMap<TreeIndex, ContractState>, Box<dyn Error>> {
```
odesenfans, line 124: Avoid returning `Box<dyn Error>`, it's not the way to go. Two solutions here:

1. Use `thiserror`, but it's pretty verbose for a binary application
2. Use `anyhow`.

I think we can use `anyhow` and get decent exception logging here.

crates/bin/prove_block/src/state_utils.rs
```diff
@@ -26,81 +40,205 @@ use crate::rpc_utils::CachedRpcStorage;
...
+    // Update contract storage roots with cached changes.
+    let empty_updates = HashMap::new();
+    let mut updated_contract_states: HashMap<num_bigint::BigUint, ContractState> = HashMap::new();
+
+    for address in processed_state_update.accessed_addresses.clone() {
```
odesenfans, line 129: No need to clone it, it's a `HashSet<Felt>` so values can be copied instead.

odesenfans (changes requested):

crates/bin/prove_block/src/state_utils.rs
HermanObst: Its a "SharedState compatible" set of updates. 
Maybe 'FormattedStateUpdate' ?

crates/bin/prove_block/src/state_utils.rs
HermanObst: done :)

crates/bin/prove_block/src/state_utils.rs
HermanObst: Ok. Use` .iter()` instead

crates/bin/prove_block/src/state_utils.rs
```diff
@@ -26,81 +40,205 @@ use crate::rpc_utils::CachedRpcStorage;
...
+        // unwrap() is safe as an entry is guaranteed to be present with `get_leaves()`.
+        let tree_index = address.to_biguint();
+        let updates = processed_state_update.storage_updates.get(&address).unwrap_or(&empty_updates);
```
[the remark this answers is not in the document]
HermanObst: I can't trivially use that as Default for HashMap<Felt, Felt> is not implemented. 
Maybe we can define a better type and define Default in another PR?
@odesenfans

crates/bin/prove_block/src/state_utils.rs
```diff
@@ -26,81 +40,205 @@ use crate::rpc_utils::CachedRpcStorage;
...
+        let nonce = processed_state_update.address_to_nonce.get(&address).cloned();
```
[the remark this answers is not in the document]
HermanObst: Isn't the other way around? If I use `copied()`, I need the inner elements to implement `Copy`?

crates/bin/prove_block/src/state_utils.rs
```diff
@@ -26,81 +40,205 @@ use crate::rpc_utils::CachedRpcStorage;
...
+        let mut class_hash = processed_state_update.address_to_class_hash.get(&address).cloned();
+        if class_hash.is_none() {
+            let resp = provider.get_class_hash_at(block_id, address).await;
+            class_hash = if let Ok(class_hash) = resp { Some(class_hash) } else { Some(Felt252::ZERO) };
+        }
+        let updated_contract_state =
+            empty_contract_states.remove(&tree_index).unwrap().update(ffc, updates, nonce, class_hash).await?;
```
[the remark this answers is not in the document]
HermanObst: @odesenfans using `empty_contract_states.get_mut(&tree_index)` will require `ContractState` to implement `Copy` and that cant be done as `contract_address: Vec<u8>`

crates/bin/prove_block/src/state_utils.rs
```diff
@@ -26,81 +40,205 @@ use crate::rpc_utils::CachedRpcStorage;
...
+    provider: &JsonRpcClient<HttpTransport>,
+    block_id: BlockId,
+    ffc: &mut FactFetchingContext<CachedRpcStorage, PedersenHash>,
+) -> Result<HashMap<TreeIndex, ContractState>, Box<dyn Error>> {
+    // Update contract storage roots with cached changes.
+    let empty_updates = HashMap::new();
+    let mut updated_contract_states: HashMap<num_bigint::BigUint, ContractState> = HashMap::new();
+
+    for address in processed_state_update.accessed_addresses.clone() {
+        // unwrap() is safe as an entry is guaranteed to be present with `get_leaves()`.
```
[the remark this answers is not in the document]
HermanObst: line 144

crates/bin/prove_block/src/state_utils.rs
odesenfans: Yeah, something like that would be clearer.

crates/bin/prove_block/src/state_utils.rs
odesenfans: Yeah that should work.

crates/bin/prove_block/src/state_utils.rs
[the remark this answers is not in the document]
odesenfans: Ok. Move the comment there?

crates/bin/prove_block/src/state_utils.rs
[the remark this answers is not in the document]
odesenfans: This compiles:

```
let o: Option<HashMap<Felt252, Felt252>> = None;
let h2 = o.unwrap_or_default();
```

The types inside the hashmap should not matter when calling `default()`.

crates/bin/prove_block/src/state_utils.rs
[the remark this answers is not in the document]
odesenfans: Yes, that's what I meant :p I want to rely on the property that we only copy small types, so this way if we ever replace it for some reason we'll be aware that there's a copy being made.

crates/bin/prove_block/src/state_utils.rs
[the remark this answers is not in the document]
odesenfans: @HermanObst this is still unresolved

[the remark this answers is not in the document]
HermanObst: Sorry I mislead you in the comment before. I was tired :P 

The real problem is that im receiving a reference:

```
let o: Option<&HashMap<Felt252, Felt252>> = None;
let h2 = o.unwrap_or_default(); // Not compile
```

Possible options:
Cleaner way than the current 
1. `let updates: &HashMap<Felt252, Felt252> = processed_state_update.storage_updates.get(address).unwrap_or(&HashMap::default());
`

2. `let updates: HashMap<Felt252, Felt252> = processed_state_update.storage_updates.get(address).cloned().unwrap_or_default();
`

I can use the want you think is better :) 

Is there something im missing?
@odesenfans

odesenfans (approved):

PR #322, opened 2024-08-23
Testing: Allow proving of blocks in CI

whichqua:
Problem: Currently there is no way to run the `prove_block` command against several blocks in the CI.
Solution: Implement a way to start pathfinder and prove blocks.

Issue Number: N/A

## Type

- [ ] feature
- [ ] bugfix
- [ ] build
- [ ] docs
- [x] testing

## Breaking changes?

- [ ] yes
- [x] no

### DIFF
--- a/.github/workflows/prove_block.yml
+++ b/.github/workflows/prove_block.yml
@@ -5,9 +5,6 @@ on:
     branches:
       - main
   pull_request:
-    paths:
-      - "crates/bin/prove_block/**"
-      - ".github/workflows/prove_block.yml"
 
 jobs:
   build:
@@ -16,50 +13,6 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
-
-      - name: Install required packages
-        run: |
-          sudo apt-get update
-          sudo apt-get install -y curl git python3 python3-venv python3-dev build-essential libgmp-dev pkg-config libssl-dev zstd
-
-      - name: Cache Db
-        id: cache-db
-        uses: actions/cache@v3
-        with:
-          path: |
-            $HOME/pathfinder-data/testnet-sepolia.sqlite
-
-      - name: Download and extract database file
-        if: ${{ steps.cache-db.outputs.cache-hit != 'true' }}
-        run: |
-          mkdir -p $HOME/pathfinder-data
-          curl -L https://pub-1fac64c3c0334cda85b45bcc02635c32.r2.dev/sepolia-testnet_0.13.0_74494_archive.sqlite.zst -o sepolia-testnet.sqlite.zst
-          zstd -d sepolia-testnet.sqlite.zst -o $HOME/pathfinder-data/testnet-sepolia.sqlite
-
-      - name: Pull Pathfinder Docker image
-        run: |
-          docker pull odesenfans/pathfinder:latest
-      - name: Run Pathfinder container in the background
-        run: |
-          docker run \
-            --name pathfinder \
-            --detach \
-            -p 127.0.0.1:9545:9545 \
-            --user "$(id -u):$(id -g)" \
-            -e RUST_LOG=info \
-            -e PATHFINDER_ETHEREUM_API_URL=${{ secrets.PATHFINDER_ETHEREUM_API_URL }} \
-            -v $HOME/pathfinder-data:/usr/share/pathfinder/data \
-            odesenfans/pathfinder:latest \
-            --storage.state-tries=archive
-      - name: Wait for Pathfinder to be ready
-        run: |
-          until curl -s http://127.0.0.1:9545; do
-            echo "Waiting for Pathfinder to start..."
-            docker logs pathfinder
-            sleep 5
-          done
-          echo "Pathfinder is up!"
-          docker logs pathfinder
 
       - name: "Install Rust"
         uses: "actions-rs/toolchain@v1"
@@ -92,4 +45,4 @@ jobs:
           bash scripts/setup-tests.sh
 
       - name: Prove Blocks
-        run: bash scripts/prove-blocks.sh -p http://127.0.0.1:9545 -b 33911,33922,33928
+        run: bash scripts/prove-blocks.sh -p ${{ secrets.PATHFINDER_RPC_URL }} -b 33911,33922,33928
--- a/.github/workflows/prove_block.yml
+++ b/.github/workflows/prove_block.yml
@@ -1,0 +1,92 @@
+name: Prove Blocks
+
+on:
+  push:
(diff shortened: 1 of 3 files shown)


.github/workflows/prove_block.yml
```diff
@@ -0,0 +1,95 @@
+name: Prove Blocks
```
odesenfans, line 1:
```suggestion
name: Sepolia integration tests
```
?

.github/workflows/prove_block.yml
```diff
@@ -0,0 +1,95 @@
...
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    paths:
+      - "crates/bin/prove_block/**"
```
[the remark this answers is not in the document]
odesenfans: Not really, if there are changes to the OS then we need to check this. For example, this action did not run for the current PR. I'd leave it on by default, let's see how much time it takes and if we need to run it less often.

.github/workflows/prove_block.yml
```diff
@@ -0,0 +1,95 @@
...
+      - ".github/workflows/prove_block.yaml"
```
odesenfans, line 10:
```suggestion
      - ".github/workflows/prove_block.yml"
```

.github/workflows/prove_block.yml
```diff
@@ -0,0 +1,95 @@
...
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
```
odesenfans, line 18: I think you'll need to sync the submodules.

.github/workflows/prove_block.yml
```diff
@@ -0,0 +1,95 @@
...
+
+      - name: Install required packages
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y curl git python3 python3-venv python3-dev build-essential libgmp-dev pkg-config libssl-dev zstd
+
+      - name: Cache Db
```
odesenfans, line 25: Does this work? I thought we were limited to 10GB.

odesenfans:
```suggestion
      - name: Cache DB
```

.github/workflows/prove_block.yml
```diff
@@ -0,0 +1,95 @@
...
+          docker pull odesenfans/pathfinder:latest
+      - name: Run Pathfinder container in the background
+        run: |
+          docker run \
+            --name pathfinder \
+            --detach \
+            -p 127.0.0.1:9545:9545 \
+            --user "$(id -u):$(id -g)" \
+            -e RUST_LOG=info \
+            -e PATHFINDER_ETHEREUM_API_URL=${{ secrets.PATHFINDER_ETHEREUM_API_URL }} \
```
odesenfans, line 50: This means that you will sync with Ethereum. I see that you can set `PATHFINDER_SYNC_ENABLED` to false to disable sync.

.github/workflows/prove_block.yml
```diff
@@ -0,0 +1,95 @@
...
+            -v $HOME/pathfinder-data:/usr/share/pathfinder/data \
+            odesenfans/pathfinder:latest \
+            --storage.state-tries=archive
+      - name: Wait for Pathfinder to be ready
+        run: |
+          until curl -s http://127.0.0.1:9545; do
+            echo "Waiting for Pathfinder to start..."
+            docker logs pathfinder
+            sleep 5
```
odesenfans, line 59: 5 seconds is long, you can sleep for 0.1 / 0.2 seconds and be more reactive. Also, this will loop forever if Pathfinder fails to start, so add a failsafe.

.github/workflows/prove_block.yml
```diff
@@ -0,0 +1,95 @@
...
+          source venv/bin/activate
+
+      - name: Setup the tests
+        run: |
+          source venv/bin/activate
+          pip install cairo-lang==0.13.1 "sympy<1.13.0"
+          bash scripts/setup-tests.sh
+
+      - name: Prove Blocks
+        run: bash scripts/prove-blocks.sh -p http://127.0.0.1:9545 -b 33911,33922,33928
```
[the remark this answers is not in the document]
odesenfans: These work?

scripts/prove-blocks.sh
```diff
@@ -0,0 +1,40 @@
...
+
+# Check if both provider and block numbers are provided
+if [ -z "$PROVIDER" ] || [ -z "$BLOCK_NUMBERS" ]; then
+  usage
+fi
+
+# Loop through each block number and execute the cargo command
+for BLOCK_NUMBER in "${BLOCK_NUMBERS[@]}"; do
+  echo "Executing for block number: $BLOCK_NUMBER"
+  cargo run -p prove_block -- --block-number "$BLOCK_NUMBER" --rpc-provider "$PROVIDER"
```
odesenfans, line 31: Run in release mode.

scripts/prove-blocks.sh
```diff
@@ -0,0 +1,40 @@
...
+  
+  # Check if the cargo command failed
+  if [ $? -ne 0 ]; then
+    echo "Command failed for block number: $BLOCK_NUMBER. Exiting."
+    exit 1
+  fi
+done
+
+echo "All blocks proved successfully."
```
odesenfans, line 40:
```suggestion
echo "All blocks proven successfully."
```

odesenfans (changes requested):

HermanObst (approved):
LGTM

scripts/prove-blocks.sh
```diff
@@ -0,0 +1,40 @@
...
+  cargo run -p prove_block --release -- --block-number "$BLOCK_NUMBER" --rpc-provider "$PROVIDER"
+  
+  # Check if the cargo command failed
+  if [ $? -ne 0 ]; then
+    echo "Command failed for block number: $BLOCK_NUMBER. Exiting."
+    exit 1
+  fi
+done
+
+echo "All blocks proven successfully."
```
odesenfans, line 40: My main question is regarding this script: why did you not implement this as a set of integration tests in Rust instead of a bash script? Of course, this works, but it feels like this could be a set of tests in the `tests` directory of `prove_block`. You just need to reorganize the main function to move the implementation inside a separate function in a library module and call it from the tests.

odesenfans: As discussed with @HermanObst , let's migrate this to Rust in another PR. We'll merge this now in order to add blocks in other PRs to spot regressions.

odesenfans (approved):

PR #325, opened 2024-08-26
Chore: remove unused code in prove block main

HermanObst:
## Description
Problem: We have logic in our main.rs that is:
- Not used at all
- Duplicated

Solution:
Remove the unused code.

⚠️ Please run prove_block with our successful block to ensure that we dont get regressions

Explanation:
```
    // fill in class hashes for each accessed contract
    for address in &accessed_contracts {
        if *address != Felt252::ONE && !address_to_class_hash.contains_key(address) {
            log::info!("Querying missing class hash for {}", address);
            let class_hash = provider.get_class_hash_at(BlockId::Number(block_number), address).await.unwrap();
            log::info!("Got class hash: {} => {}", address, class_hash);
            address_to_class_hash.insert(*address, class_hash);
        }
    }
```
This is handled in https://github.com/keep-starknet-strange/snos/pull/321/files#diff-4e383172600a3ec792a28a49a94d101c9f2a7dc65d2e0052b386ceafe3fd509eR240-R258

```
    // TODO: clean up...
    // now that we have the class_hash for each contract, fill in the transaction data with the
    // class_hash. when transactions were first processed above, this information wasn't available.
    for transaction in transactions.iter_mut() {
        if let Some(sender_address) = transaction.sender_address {
            let class_hash = address_to_class_hash
                .get(&sender_address)
                .expect("should have a class_hash for each known contract addresses at this point");
            transaction.class_hash = Some(*class_hash);
        } else {
            // TODO: are there txn types which wouldn't have a sender address?
            unimplemented!("Found transaction without sender_address");
        }
    }
```
Is just not used

## Breaking changes?

- [ ] yes
- [X] no

### DIFF
--- a/crates/bin/prove_block/src/main.rs
+++ b/crates/bin/prove_block/src/main.rs
@@ -11,7 +11,7 @@ use rpc_replay::block_context::build_block_context;
 use rpc_replay::rpc_state_reader::AsyncRpcStateReader;
 use rpc_replay::transactions::starknet_rs_to_blockifier;
 use rpc_utils::{get_class_proofs, get_storage_proofs, TrieNode};
-use starknet::core::types::{BlockId, MaybePendingBlockWithTxs, MaybePendingBlockWithTxHashes};
+use starknet::core::types::{BlockId, MaybePendingBlockWithTxHashes, MaybePendingBlockWithTxs};
 use starknet::providers::jsonrpc::HttpTransport;
 use starknet::providers::{JsonRpcClient, Provider, Url};
 use starknet_os::config::{StarknetGeneralConfig, StarknetOsConfig, SN_SEPOLIA, STORED_BLOCK_HASH_BUFFER};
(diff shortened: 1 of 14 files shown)


crates/bin/prove_block/src/main.rs
```diff
@@ -124,21 +124,26 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
-            MaybePendingBlockWithTxs::PendingBlock(_) => {
-                panic!("Block is still pending!");
-            }
-        };
+
+    // We only need to get the older block number and hash. No need to fetch all the txs
+    let older_block = match provider
+        .get_block_with_tx_hashes(BlockId::Number(block_number - STORED_BLOCK_HASH_BUFFER))
+        .await
+        .unwrap()
```
odesenfans, line 132: For another PR: now that we use `anyhow`, these unwraps should be replaced by `.with_context(|| format!("failed to retrieve previous block ({})", previous_block_id))`

odesenfans (approved):

PR #327, opened 2024-08-28
Chore: remove build_initial_state

HermanObst:
## Description
Problem: all the code related with build_initial_state was useful
Solution: remove it

## Breaking changes?

- [ ] yes
- [X] no

### DIFF
--- a/crates/bin/prove_block/src/state_utils.rs
+++ b/crates/bin/prove_block/src/state_utils.rs
@@ -1,151 +1,24 @@
 use std::collections::{HashMap, HashSet};
 
-use anyhow::Result;
 use cairo_vm::Felt252;
-use num_bigint::BigUint;
 use starknet::core::types::{BlockId, MaybePendingStateUpdate, StateDiff};
 use starknet::providers::jsonrpc::HttpTransport;
 use starknet::providers::{JsonRpcClient, Provider};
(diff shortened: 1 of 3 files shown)


crates/bin/prove_block/src/state_utils.rs
```diff
@@ -1,153 +1,26 @@
...
-use starknet_os::starkware_utils::commitment_tree::patricia_tree::patricia_tree::PatriciaTree;
-use starknet_os::storage::storage::FactFetchingContext;
 
-use crate::rpc_utils::{CachedRpcStorage, RpcStorage};
 use crate::utils::get_subcalled_contracts_from_tx_traces;
 
 #[derive(Clone)]
 pub struct FormattedStateUpdate {
     // TODO: Use more descriptive types
+    #[allow(unused)]
```
odesenfans, line 14: These are pretty suspicious. If they're unused, why do we keep them?

odesenfans (commented):
One question. Good!

crates/bin/prove_block/src/state_utils.rs
HermanObst: This is part of the TODOs (Im going to need to touch more stuff to remove them) 
But they are also gone in the next PR :P

odesenfans (approved):
LGTM

PR #328, opened 2024-08-28
Feature: support Sierra 1.6

odesenfans:
Problem: some contracts on Sepolia are built with Sierra 1.6, which is
not supported by the version of the CASM compiler we use. But bumping
the version of the compiler requires bumping cairo-vm and Blockifier as
well.

Solution: bump everything! We now use:
* cairo-vm 1.0.1
* cairo-lang 2.7.1
* blockifier 0.8.0-rc.2

With this, we can now compile any contract on Sepolia.

Issue Number: N/A

## Type

- [ ] feature
- [ ] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [ ] yes
- [ ] no

### DIFF
--- a/crates/rpc-replay/src/block_context.rs
+++ b/crates/rpc-replay/src/block_context.rs
@@ -1,15 +1,19 @@
-use blockifier::block::{BlockInfo, GasPrices};
+use blockifier::blockifier::block::{BlockInfo, GasPrices};
+use blockifier::bouncer::BouncerConfig;
 use blockifier::context::{BlockContext, ChainInfo, FeeTokenAddresses};
 use blockifier::versioned_constants::VersionedConstants;
 use starknet::core::types::BlockWithTxs;
 use starknet_api::block::{BlockNumber, BlockTimestamp};
-use starknet_api::core::{ContractAddress, PatriciaKey};
-use starknet_api::hash::StarkHash;
-use starknet_api::{contract_address, patricia_key};
+use starknet_api::core::{ChainId, ContractAddress, PatriciaKey};
+use starknet_api::{contract_address, felt, patricia_key};
 
 use crate::utils::felt_to_u128;
 
-pub fn build_block_context(chain_id: String, block: &BlockWithTxs) -> BlockContext {
+pub fn build_block_context(
+    chain_id: ChainId,
+    block: &BlockWithTxs,
+    starknet_version: blockifier::versioned_constants::StarknetVersion,
+) -> BlockContext {
     let sequencer_address_hex = block.sequencer_address.to_hex_string();
     let sequencer_address = contract_address!(sequencer_address_hex.as_str());
 
@@ -27,7 +31,7 @@ pub fn build_block_context(chain_id: String, block: &BlockWithTxs) -> BlockConte
     };
 
     let chain_info = ChainInfo {
-        chain_id: starknet_api::core::ChainId(chain_id),
+        chain_id,
         // cf. https://docs.starknet.io/tools/important-addresses/
         fee_token_addresses: FeeTokenAddresses {
             strk_fee_token_address: contract_address!(
@@ -39,7 +43,8 @@ pub fn build_block_context(chain_id: String, block: &BlockWithTxs) -> BlockConte
         },
     };
 
-    let versioned_constants = VersionedConstants::latest_constants();
+    let versioned_constants = VersionedConstants::get(starknet_version);
+    let bouncer_config = BouncerConfig::max();
 
-    BlockContext::new_unchecked(&block_info, &chain_info, versioned_constants)
+    BlockContext::new(block_info, chain_info, versioned_constants.clone(), bouncer_config)
 }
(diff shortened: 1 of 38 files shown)


crates/starknet-os-types/src/chain_id.rs
```diff
@@ -5,3 +5,21 @@ use starknet_types_core::felt::Felt;
 pub fn chain_id_to_felt(chain_id: &ChainId) -> Felt {
     Felt::from_bytes_be_slice(chain_id.to_string().as_bytes())
 }
+
+pub fn chain_id_from_felt(felt: Felt) -> ChainId {
+    let chain_id_bytes: Vec<_> = felt.to_bytes_be().into_iter().skip_while(|byte| *byte == 0u8).collect();
```
HermanObst, line 10: Can you comment the motivation of this?
@odesenfans

crates/rpc-replay/src/block_context.rs
```diff
@@ -1,15 +1,19 @@
-use blockifier::block::{BlockInfo, GasPrices};
+use blockifier::blockifier::block::{BlockInfo, GasPrices};
+use blockifier::bouncer::BouncerConfig;
```
HermanObst, line 2: What is a Bouncer?

odesenfans: If I understand it right, it's a mechanism used to determine how many txs can go in a block depending on different factors like gas, builtin usage, number of events etc.

crates/starknet-os/src/execution/deprecated_syscall_handler.rs
```diff
@@ -273,17 +263,12 @@ where
...
         Ok(())
     }
     pub async fn storage_write(&self, _syscall_ptr: Relocatable) -> Result<(), HintError> {
-        let sys_hand = self.deprecated_syscall_handler.write().await;
-
-        let _ = sys_hand.exec_wrapper.execution_helper.write().await.execute_code_read_iter.next().ok_or(
-            HintError::SyscallError("No more storage writes available to replay".to_string().into_boxed_str()),
-        )?;
-
+        // Nothing to do
```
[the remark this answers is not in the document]
odesenfans: Good remark! Newer versions of Blockifier now split reads and writes into two separate fields, `CallInfo.storage_read_values` and `CallInfo.accessed_storage_keys`. Before the two were merged so we had to take a value from the iterator (writes were always set to 0, the value itself has no interest) so that the read syscall could take the correct values.

Now, we can just iterate over `storage_read_values` in the read syscall and do nothing in the write syscall. Technically we could iterate over the `accessed_storage_keys` in the write syscall but there is no interest in doing that besides checking that we have an equal amount of values. It sounds a bit useless to me so I just replaced this code by a no-op.

crates/starknet-os/src/storage/storage_utils.rs
```diff
@@ -68,7 +67,7 @@ pub async fn unpack_blockifier_state_async<S: Storage + Send + Sync, H: HashFunc
 ) -> Result<(SharedState<S, H>, SharedState<S, H>), TreeError> {
     let final_state = {
         let state = blockifier_state.state.clone();
-        state.apply_commitment_state_diff(blockifier_state.to_state_diff()).await?
+        state.apply_commitment_state_diff(blockifier_state.to_state_diff().expect("todo").into()).await?
```
HermanObst, line 70: `.expect("todo")` seems vague :P

crates/starknet-os/src/starknet/business_logic/fact_state/state.rs
```diff
@@ -430,22 +425,20 @@ where
...
-    async fn get_storage_at_async(
-        &mut self,
-        contract_address: ContractAddress,
-        key: StorageKey,
-    ) -> StateResult<StarkFelt> {
-        let storage_key: TreeIndex = felt_api2vm(*key.0.key()).to_biguint();
+    async fn get_storage_at_async(&self, contract_address: ContractAddress, key: StorageKey) -> StateResult<Felt252> {
+        let storage_key: TreeIndex = key.0.key().to_biguint();
+
+        let mut ffc = self.ffc.clone();
```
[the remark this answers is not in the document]
odesenfans: This is because the trait now forces this method to take `&self` instead of `&mut self`. The FFC is built around an `Arc<Mutex<Storage>>` so we can just deal with this problem by cloning the FFC and then modifying the clone.

crates/starknet-os/src/storage/storage_utils.rs
odesenfans: Woops, good point.

odesenfans: Fixed. This function should not be there btw, this is only used in one integration test. I modified the expect message to be clearer, I'll move the function in another PR.

HermanObst (approved):
Awesome job! LGTM!

crates/starknet-os/src/storage/storage_utils.rs
HermanObst: Nice! Thanks!

PR #330, opened 2024-08-29
Feature: handle all tx traces

ftheirs:
Problem: at the moment only Invoke transactions were supported

Solution: process traces and add type conversion for reverted Invoke transaction, DeployAccount, Declare and L1Handler

Issue Number: N/A

## Type

- [x] feature
- [ ] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [ ] yes
- [x] no

### DIFF
--- a/crates/rpc-replay/src/transactions.rs
+++ b/crates/rpc-replay/src/transactions.rs
@@ -41,7 +41,7 @@ fn da_mode_core_to_api(
 pub fn starknet_rs_to_blockifier(
     sn_core_tx: &starknet::core::types::Transaction,
 ) -> Result<blockifier::transaction::transaction_execution::Transaction, Box<dyn Error>> {
-    let blockifier_tx: AccountTransaction = match sn_core_tx {
+    let blockifier_tx = match sn_core_tx {
         Transaction::Invoke(tx) => {
             let (tx_hash, api_tx) = match tx {
                 InvokeTransaction::V0(tx) => {
@@ -90,9 +90,13 @@ pub fn starknet_rs_to_blockifier(
                     (tx_hash, api_tx)
                 }
             };
+
             let invoke =
                 blockifier::transaction::transactions::InvokeTransaction { tx: api_tx, tx_hash, only_query: false };
-            AccountTransaction::Invoke(invoke)
+
+            blockifier::transaction::transaction_execution::Transaction::AccountTransaction(AccountTransaction::Invoke(
+                invoke,
+            ))
         }
         Transaction::DeployAccount(_tx) => {
             unimplemented!("starknet_rs_tx_to_blockifier() with Deploy txn");
@@ -106,5 +110,5 @@ pub fn starknet_rs_to_blockifier(
         _ => unimplemented!(),
     };
 
-    Ok(blockifier::transaction::transaction_execution::Transaction::AccountTransaction(blockifier_tx))
+    Ok(blockifier_tx)
 }
--- a/crates/rpc-replay/src/transactions.rs
+++ b/crates/rpc-replay/src/transactions.rs
@@ -54,7 +54,7 @@ fn invoke_v1_to_blockifier(
         max_fee: starknet_api::transaction::Fee(tx.max_fee.to_biguint().try_into()?),
         signature: starknet_api::transaction::TransactionSignature(tx.signature.to_vec()),
         nonce: starknet_api::core::Nonce(tx.nonce),
-        sender_address: starknet_api::core::ContractAddress(PatriciaKey::try_from(tx.sender_address).unwrap()),
+        sender_address: starknet_api::core::ContractAddress(PatriciaKey::try_from(tx.sender_address)?),
         calldata: starknet_api::transaction::Calldata(Arc::new(tx.calldata.to_vec())),
     });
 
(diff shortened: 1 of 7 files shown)


HermanObst (changes requested):
@ftheirs please modularize all the different tx with each version.
see 
[starknet_rs_tx_to_internal_tx](https://github.com/keep-starknet-strange/snos/blob/main/crates/bin/prove_block/src/types.rs#L186-L205)

crates/rpc-replay/src/transactions.rs
```diff
@@ -38,10 +44,12 @@ fn da_mode_core_to_api(
 }
 
 /// Maps starknet-core transactions to Blockifier-compatible types.
-pub fn starknet_rs_to_blockifier(
+pub async fn starknet_rs_to_blockifier(
     sn_core_tx: &starknet::core::types::Transaction,
+    provider: &JsonRpcClient<HttpTransport>,
+    block_number: u64,
 ) -> Result<blockifier::transaction::transaction_execution::Transaction, Box<dyn Error>> {
```
odesenfans, line 51: For another PR, but we really need to return something better than `Box<dyn Error>` here.

crates/rpc-replay/src/transactions.rs
```diff
@@ -90,21 +98,196 @@ pub fn starknet_rs_to_blockifier(
...
+                    );
+
+                    (tx_hash, api_tx, tx.class_hash)
+                }
+
+                _ => unimplemented!("DeclareTransaction V0 not supported"),
+            };
+
+            // TODO: improve this to avoid retrieving this twice. Already done in main.rs from prove_block
+            let starknet_contract_class = provider.get_class(BlockId::Number(block_number), class_hash).await?;
```
odesenfans, line 242: Could you put the `ClassInfo` generation in a new function?

crates/rpc-replay/src/transactions.rs
```diff
@@ -90,21 +98,196 @@ pub fn starknet_rs_to_blockifier(
...
+                starknet::core::types::ContractClass::Sierra(flattened_sierra_cc) => {
+                    GenericSierraContractClass::from(flattened_sierra_cc)
+                }
+                starknet::core::types::ContractClass::Legacy(_) => {
+                    unimplemented!("Fixme: Support legacy contract class")
+                }
+            };
+
+            let flattened_sierra = generic_sierra_cc.clone().to_starknet_core_contract_class()?;
+            let contract_class = generic_sierra_cc.compile()?.get_blockifier_contract_class()?.clone();
```
odesenfans, line 253: There's (or at least should be) a `to_blockifier_contract_class` method that hides the clone.

crates/rpc-replay/src/transactions.rs
```diff
@@ -90,21 +98,196 @@ pub fn starknet_rs_to_blockifier(
...
+
+                    (tx_hash, api_tx, tx.class_hash)
+                }
+
+                _ => unimplemented!("DeclareTransaction V0 not supported"),
+            };
+
+            // TODO: improve this to avoid retrieving this twice. Already done in main.rs from prove_block
+            let starknet_contract_class = provider.get_class(BlockId::Number(block_number), class_hash).await?;
+            let generic_sierra_cc = match starknet_contract_class {
```
odesenfans, line 243: To deal with the legacy/deprecated/Cairo0 (these are synonyms) contracts, you simply need to move the compilation code + conversion to Blockifier `ContractClass` here and do the appropriate thing for Cairo0 in the other arm of the match. Please resolve it in this PR.

crates/rpc-replay/src/transactions.rs
```diff
@@ -90,21 +98,196 @@ pub fn starknet_rs_to_blockifier(
...
+
+            let declare = blockifier::transaction::transactions::DeclareTransaction::new(api_tx, tx_hash, class_info)?;
+
+            blockifier::transaction::transaction_execution::Transaction::AccountTransaction(
+                AccountTransaction::Declare(declare),
+            )
         }
-        Transaction::L1Handler(_tx) => {
-            unimplemented!("starknet_rs_tx_to_blockifier() with L1Handler txn");
+        Transaction::L1Handler(tx) => {
```
odesenfans, line 267: As @HermanObst already suggested, please split every arm of this match into separate functions.

crates/rpc-replay/src/transactions.rs
```diff
@@ -90,21 +98,196 @@ pub fn starknet_rs_to_blockifier(
...
+                nonce: starknet_api::core::Nonce(Felt::from(tx.nonce)),
+                contract_address: starknet_api::core::ContractAddress(
+                    PatriciaKey::try_from(tx.contract_address).unwrap(),
+                ),
+                entry_point_selector: starknet_api::core::EntryPointSelector(tx.entry_point_selector),
+                calldata: starknet_api::transaction::Calldata(Arc::new(tx.calldata.clone())),
+            };
+
+            // TODO: Check Fee value
+            let fee = Fee(0);
```
odesenfans, line 280: Looking at the explorer, this seems to be correct. Replace the TODO by a comment saying that fees are 0 for L1 handler txs.

odesenfans (changes requested):

HermanObst (changes requested):
@ftheirs good job man! 
one last comment: remove all the `unwrap()` that you are using in favor to correctly handle the errors.

In general, we never accept unwrap() as they can make the program panic, basically bypassing Rust safety.
We only accept if there is a comment, clearly explaining why a given unwrap() will never fail. But handling the error is better practice. 

 Sorry that I didnt catch it before

ftheirs:
Closing this since it will be split in several PRs

PR #339, opened 2024-09-02
Fix: support `replace_class()` calls in invoke txs

odesenfans:
Problem: some blocks fail on divergences in the global state tree caused by calls to the `replace_class()` syscall.

Solution: use the class hash in block N-1 when building the `ContractClass` objects instead of the current block, and fetch the compiled class for both blocks to build the trees properly.

Issue Number: N/A

## Type

- [ ] feature
- [x] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [ ] yes
- [x] no


HermanObst (approved):
Nicely done! 
LGTM!

PR #340, opened 2024-09-03
Fix: Impl missing hints in Blob DA blocks

notlesh:
Problem: SNOS is missing some hints related to BLOB transactions

Solution: Add the missing hints

Issue Number: N/A

## Type

- [ ] feature
- [x] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [ ] yes
- [x] no


notlesh:
(Actually, this PR may not be needed after all...)

PR #341, opened 2024-09-03
Fix: support writes to zero

odesenfans:
Problem: we encounter an "Edge bottom not found in preimage" error when running some blocks. This problem was diagnosed as writes to zero where the last node of the storage proof was an edge node. A write to zero to a specific address will clear the path to that leaf node, so the storage proof for this key will always be incomplete. It turns out that the OS will still look for the bottom node in the preimage to verify the node hash.

Solution: if we encounter such a case, we now fetch an additional key that is designed to follow the edge path. This way, we get the bottom node in the preimage and the OS can compute the hash it wants to compute.

Issue Number: N/A

## Type

- [ ] feature
- [x] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [ ] yes
- [x] no


crates/bin/prove_block/src/rpc_utils.rs
```diff
@@ -150,40 +249,35 @@ pub(crate) async fn get_storage_proofs(
...
-fn merge_chunked_storage_proofs(proofs: Vec<PathfinderProof>) -> PathfinderProof {
+/// Returns a modified key that follows the specified edge path.
+/// This function is used to work around an issue where the OS fails if it encounters a
+/// write to 0 and the last node in the storage proof is an edge node of length 1.
+/// In this situation the OS will still look up the node in the preimage and will fail
+/// on a "Edge bottom not found in preimage" error.
+/// To resolve this, we fetch the storage proof for a node that follows this edge in order
+/// to get the bottom node in the preimage and resolve the issue.
+fn get_key_following_edge(key: Felt, height: Height, edge_path: &EdgePath) -> Felt {
+    let mask = edge_path.value.to_bigint() << (251 - height.0);
```
notlesh, line 274: Could use a constant instead of `251`

notlesh: Maybe also assert that `height.0 < 251`?

notlesh (approved):
Looks good other than a couple of cosmetic suggestions.

Another question, though: is there any equivalent of this for the class tree?

odesenfans:
> Looks good other than a couple of cosmetic suggestions.
> 
> Another question, though: is there any equivalent of this for the class tree?

There is no equivalent yet, I don't think this is necessary. There is no way to delete a class, so there is no "write to 0" for the contract and class tries. Furthermore, you are always guaranteed to always go 0 -> something or X -> Y, so there will always be a valid proof for at least the current state.

PR #343, opened 2024-09-03
clean up Invoke tx

ftheirs:
Problem PR https://github.com/keep-starknet-strange/snos/pull/330 was too big

Solution: split the PR into smaller PR where each one will implement a different transaction. 
Clean up and modularize Invoke tx

Issue Number: N/A

## Type

- [x] feature
- [ ] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [ ] yes
- [x] no

### DIFF
--- a/crates/rpc-replay/src/transactions.rs
+++ b/crates/rpc-replay/src/transactions.rs
@@ -3,9 +3,12 @@ use std::error::Error;
 use std::sync::Arc;
 
 use blockifier::transaction::account_transaction::AccountTransaction;
-use starknet::core::types::{InvokeTransaction, ResourceBoundsMapping, Transaction};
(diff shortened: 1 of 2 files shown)


crates/rpc-replay/src/transactions.rs
```diff
@@ -37,63 +40,60 @@ fn da_mode_core_to_api(
     }
 }
 
+fn felt_to_fee(fee: Felt) -> Result<Fee, StarknetApiError> {
+    let fee_u128 = fee
```
odesenfans, line 44: We have a `felt_to_u128` utility already, can you use that one instead?

odesenfans (approved):

PR #346, opened 2024-09-04
Fix: support invoke txs that call the deploy syscall

odesenfans:
Problem: the setup for `prove_block` assumes that we are always able to fetch the class hash, class and storage proof for the contract addresses found in invoke txs in the previous block. This is not always the case as invoke txs are able to deploy contracts.

Solution: accept "contract not found" errors and storage proofs where the `contract_data` field is missing. Note that we should implement a way to double-check our expectations as this fix will accept an incomplete storage proof in the current block as well, which is not ideal.

Issue Number: N/A

## Type

- [x] feature
- [x] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [ ] yes
- [x] no

### DIFF
--- a/crates/bin/prove_block/src/lib.rs
+++ b/crates/bin/prove_block/src/lib.rs
@@ -10,7 +10,7 @@ use rpc_replay::block_context::build_block_context;
 use rpc_replay::rpc_state_reader::AsyncRpcStateReader;
 use rpc_replay::transactions::starknet_rs_to_blockifier;
 use rpc_utils::{get_class_proofs, get_storage_proofs};
-use starknet::core::types::{BlockId, MaybePendingBlockWithTxHashes, MaybePendingBlockWithTxs};
+use starknet::core::types::{BlockId, MaybePendingBlockWithTxHashes, MaybePendingBlockWithTxs, StarknetError};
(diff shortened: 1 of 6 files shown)


HermanObst (approved):
LGTM
Still, I really dont like the idea of bypassing "bad" storage proofs. I imagine this could lead to some nasty bugs later. 

What could be the different options to detect declared addresses and remove them from the ones being called for previous storage proofs?

@odesenfans @notlesh

odesenfans:
> LGTM Still, I really dont like the idea of bypassing "bad" storage proofs. I imagine this could lead to some nasty bugs later.
> 
> What could be the different options to detect declared addresses and remove them from the ones being called for previous storage proofs?
> 
> @odesenfans @notlesh

Instead of passing a simple list of contract addresses, we need to have more information about each contract/class while fetching the storage proofs. I think it would make sense to have some sort of traceability for each contract address, ex: "contract A was deployed in this block, contract B was called from this transaction" etc. Basically, we need enough info to determine where the contract address is coming from and what we need to do with it.

I'll fix another bug then try to come up with a meaningful fix.

PR #350, opened 2024-09-05
Feature: rpc-client crate

odesenfans:
Problems:
* We manage two types of clients, one for Starknet RPC endpoints and another for Pathfinder proof endpoints.
* We end up spawning way more clients than necessary because the starknet-rs client type does not implement Clone.

Solution: encapsulate all RPC-related objects in a new struct, `RpcClient`. This struct lives in the new `rpc-client` crate.

* Moved all the code related to proofs in the `rpc-client` crate.
* Updated `rpc-replay` and `prove-block` to use the new RPC client.
* Added a layer to encapsulate the raw `reqwest` client used for Pathfinder proof endpoints in the new `PathfinderRpcClient` struct.

At this stage `RpcClient` still provides direct access to the Starknet and Pathfinder RPC clients, and does not implement anything fancy (read: caching). The goal of this PR is to clean up the code and encapsulate functionalities.

Issue Number: N/A

## Type

- [x] feature
- [ ] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [ ] yes
- [x] no

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,9 +5,10 @@ members = [
     "crates/bin/hint_tool",
     "crates/bin/prove_block",
     "crates/cairo-type-derive",
+    "crates/rpc-client",
+    "crates/rpc-replay",
     "crates/starknet-os",
     "crates/starknet-os-types",
-    "crates/rpc-replay",
     "tests",
 ]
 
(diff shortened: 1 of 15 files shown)


notlesh (approved):

HermanObst (approved):
Looking good!

PR #353, opened 2024-09-05
WIP: script to run multiple blocks

odesenfans:
Issue Number: N/A

## Type

- [ ] feature
- [ ] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [ ] yes
- [ ] no

### DIFF
--- a/prove-seq-blocks.sh
+++ b/prove-seq-blocks.sh
@@ -1,0 +1,35 @@
+#!/bin/bash
+
+# Check if the correct number of arguments is provided
+if [ "$#" -ne 2 ]; then
+    echo "Usage: $0 <starting_id> <n_runs>"
+    exit 1
+fi
+
+PROGRAM="cargo run --package prove_block --release --bin prove_block -- --block-number"
+START_ID=$1
+RUNS=$2
+FAILURES=0
+
+# Create a directory to store output files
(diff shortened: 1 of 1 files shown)


prove-seq-blocks.sh
```diff
@@ -0,0 +1,35 @@
...
+    exit 1
+fi
+
+PROGRAM="cargo run --package prove_block --release --bin prove_block -- --block-number"
+START_ID=$1
+RUNS=$2
+FAILURES=0
+
+# Create a directory to store output files
+OUTPUT_DIR="output_files"
```
notlesh, line 15: This could be more descriptive

prove-seq-blocks.sh
```diff
@@ -0,0 +1,35 @@
...
+mkdir -p "$OUTPUT_DIR"
+export RPC_ENDPOINT_TESTNET="http://localhost:9545"
```
notlesh, line 17: This should be configurable (I don't use `localhost`)

notlesh:
#357 could be useful here, for example you could `grep "inner_exc error"` when `$? -ne 0`.

notlesh:
Another useful thing to print here would be the number of transactions (or something correlated to the time it takes to execute the block / reproduce the failure).

PR #362, opened 2024-09-09
Feature: support proving blocks with `DEPLOY_ACCOUNT` transactions

### LINKED ISSUE #356: feat: support `DeployAccount` transaction
Describe the Feature Request
Snos panics on blocks with Deploy account transactions. In order to achieve this, several parts of the code need to handle/implement what is required to process these transactions

Branch WIP: https://github.com/keep-starknet-strange/snos/tree/gm/handle-deploy-tx-trace

Related Code

Additional Context
On the comments, the different issues that will be described and link the PR/commit from the fix

whichqua:
Problem: Proving blocks with `DEPLOY_ACCOUNT` is not supported.

Solution: Implement for both V1 and V3

Issue Number: Fixes #356 

## Type

- [x] feature
- [ ] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [ ] yes
- [x] no

### DIFF
--- a/crates/rpc-replay/src/transactions.rs
+++ b/crates/rpc-replay/src/transactions.rs
@@ -1,19 +1,43 @@
 use std::collections::BTreeMap;
-use std::error::Error;
 use std::sync::Arc;
 
 use blockifier::blockifier::block::GasPrices;
+use blockifier::execution::contract_class::ClassInfo;
 use blockifier::transaction::account_transaction::AccountTransaction;
+use blockifier::transaction::errors::TransactionExecutionError;
+use rpc_client::RpcClient;
 use starknet::core::types::{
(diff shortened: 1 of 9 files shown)


crates/rpc-replay/src/transactions.rs
```diff
@@ -122,6 +123,92 @@ fn l1_handler_to_blockifier(
     Ok(blockifier::transaction::transaction_execution::Transaction::L1HandlerTransaction(l1_handler))
 }
 
+fn create_contract_address(
```
notlesh, line 126: `create` seems a bit vague here. Maybe `derive` would be better. Also, if there's any documentation you could link here, it would help

notlesh: or `calculate_contract_address_from_deploy_tx` (a bit obnoxious, but very clear...)

notlesh (approved):

PR #365, opened 2024-09-11
Fix: invalid type conversions in GenericSierraContractClass

odesenfans:
Problem: for a block with a declare tx, we fail when attempting to convert the flattened Sierra class into a `cairo-lang` class before compiling it.

Solution: we now store a `SierraClass` instead of a `FlattenedSierraClass` in the generic struct. This simplifies JSON conversions.

Issue Number: N/A

## Type

- [ ] feature
- [x] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [ ] yes
- [x] no


odesenfans:
This will be fixed in a better way in #367.

PR #369, opened 2024-09-12
Feat: Package compiled OS as gzip

notlesh:
### Problem:

The compiled OS JSON is around 40M in size, and this gets absorbed as-is into the binary as a result of `include_bytes(...)`. In addition, we didn't include this in our source tree, requiring downstream users to figure out how to build it.

### Solution:

Gzip the OS JSON (this gets us from roughly 40M to 1M) and include this in the binary instead, then decompress it during runtime. Also, include this gzipped JSON so users will not have to worry about it.

### The bad:

The downside of this is that we must decompress it at runtime, which looks to add about `0.07sec` at runtime on my machine.

Additionally, including the compiled OS artifact in our source tree is less ideal than having a robust build system generate it as needed.

Issue Number: N/A

## Type

- [x] feature
- [ ] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [ ] yes
- [x] no


crates/bin/prove_block/src/lib.rs
```diff
@@ -318,7 +318,8 @@ pub async fn prove_block(
         (old_block_number, old_block_hash),
     );
 
-    Ok(run_os(config::DEFAULT_COMPILED_OS, layout, os_input, block_context, execution_helper)?)
+    let uncompressed_os = config::gunzip_default_os().expect("Could not uncrompress default OS");
```
odesenfans, line 321:
```suggestion
    let uncompressed_os = config::gunzip_default_os().expect("Could not uncompress default OS");
```

tests/integration/common/transaction_utils.rs
```diff
@@ -886,7 +886,8 @@ where
         execute_txs(state, &block_context, txs, deprecated_contract_classes, contract_classes).await;
 
     let layout = config::default_layout();
-    let result = run_os(config::DEFAULT_COMPILED_OS, layout, os_input, block_context, execution_helper);
+    let uncompressed_os = config::gunzip_default_os().expect("Could not uncrompress default OS");
```
odesenfans, line 889:
```suggestion
    let uncompressed_os = config::gunzip_default_os().expect("Could not uncompress default OS");
```

odesenfans (changes requested):

crates/starknet-os/src/config.rs
```diff
@@ -67,6 +69,24 @@ const fn default_use_kzg_da() -> bool {
     true
 }
 
+/// un-gzips the default OS that is baked into the binary.
+///
+/// Uncompressed, this file is roughly 40M, and compresses effectively down to less than 1M.
+/// So it is included in the binary in gzipped form with this fn provided to reinflate it.
+///
+/// NOTE: the results are not cached, so this will always be an expensive call.
+pub fn gunzip_default_os() -> std::io::Result<String> {
```
shamsasari, line 78: Shouldn't this return `Vec<u8>` to mirror the old `DEFAULT_COMPILED_OS` and the fact `run_os` types in bytes?

scripts/teardown-tests.sh
```diff
@@ -10,3 +10,4 @@ git submodule update --init
 echo -e "\nremoving compiled contracts/programs...\n"
 rm starkware
 rm -rf build/!(os_latest.json)
```
shamsasari, line 12: This doesn't exist anymore right?

notlesh: Yes, I had the same thought but I hesitate to remove it, mostly because the whole thing is still a bit sloppy and if `os_latest.json` is left around it will consume a lot of disk space.

I could easily be persuaded to improve this if you have suggestions :)

crates/starknet-os/src/config.rs
notlesh: Good idea, added in [2c4bf97](https://github.com/keep-starknet-strange/snos/pull/369/commits/2c4bf972a49748b084e4bc99b22c7dd39fc94746)

scripts/teardown-tests.sh
notlesh: Ok, it was good to review `extglob` ("extended globbing") and the `!` operator, because this isn't what we want.

The `!` operator inverts the match, so this is effectively removing:

* everything but `os_latest.json`
* everything but `os_latest.json.gz`

Which is the same as removing everything. What we want is only the latter.

notlesh: Fixed in c1bb7be

shamsasari:
This PR is now stagnating. Either update it or close if it's no longer going to be done.

notlesh:
Closing this out as we will probably take a different approach to packaging the OS.

PR #370, opened 2024-09-12
Fix: bitmask issue in `get_key_following_edge`

odesenfans:
Problem: there is a bug in this function causing us to miss some edge bottom nodes, resulting in "Edge not found in preimage" bugs.

Solution: the bitmask needs to clear all the bits corresponding to the edge, which can only be achieved with an `& 0x0` type of operation (and not `& ~mask`. Additionally, the height was inverted compared to the rest of the OS (0 is supposed to be the height of leaf nodes, not the height of the root), this PR fixes this.

Issue Number: N/A

## Type

- [ ] feature
- [x] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [ ] yes
- [x] no


HermanObst (approved):
LGTM

PR #371, opened 2024-09-13
Fix: dest_ptr is not a relocatable

odesenfans:
Problem: there is a bug in cairo-vm that causes the `dest_ptr` variable in `segment_arena.cairo` to be set to 0 (felt) instead of a relocatable.

Solution: add a workaround that computes the correct value instead of getting it from the Cairo code.

Issue Number: N/A

## Type

- [ ] feature
- [x] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [ ] yes
- [x] no


crates/starknet-os/src/hints/execution.rs
```diff
@@ -1264,7 +1264,21 @@ pub fn add_relocation_rule(
...
+    // Bugfix: this is a workaround for an issue in cairo-vm related to the way it computes
+    // `dest_ptr` in `segment_arena.cairo`. For some reason it sets `dest_ptr` to 0 (felt) instead
+    // of a valid relocatable value.
+    // As this hint is used in other places, we need to determine if we are in `segments_arena`
+    // first. We check this by determining if the "infos" variable is declared.
+    // If it is the case, we simply recompute `dest_ptr` manually.
+
+    let dest_ptr = if let Ok(infos) = get_ptr_from_var_name(vars::ids::INFOS, vm, ids_data, ap_tracking) {
+        let infos_0_end = vm.get_relocatable((infos + 1)?)?;
+        (infos_0_end + 1u32)?
```
notlesh, line 1277:
```suggestion
        let infos_0_end = vm.get_relocatable((infos + 1)?)?;
        log::warn!("rewriting dest_ptr for segment_arena workaround");
        (infos_0_end + 1u32)?
```

notlesh: A warning seems appropriate here in case this has unintended side effects...

notlesh (approved):

HermanObst (approved):
LGTM
Still, we would need to learn more about the issue in cairo-vm and request a fix

PR #375, opened 2024-09-18
Added tests for `fill_holes_in_rc96_segment` and `search_sorted_optimistic` hints

shamsasari:
`search_sorted_optimistic` was previously called `find_element` and has been renamed to match the correct Cairo function it's called from, since there are three variations of very similar hint code.

### DIFF
--- a/crates/starknet-os/src/hints/find_element.rs
+++ b/crates/starknet-os/src/hints/find_element.rs
@@ -17,7 +17,7 @@ use crate::utils::get_variable_from_root_exec_scope;
 
 #[rustfmt::skip]
 
-pub const FIND_ELEMENT: &str = indoc! {r#"array_ptr = ids.array_ptr
+pub const SEARCH_SORTED_OPTIMISTIC: &str = indoc! {r#"array_ptr = ids.array_ptr
 elm_size = ids.elm_size
 assert isinstance(elm_size, int) and elm_size > 0, \
     f'Invalid value for elm_size. Got: {elm_size}.'
@@ -39,7 +39,7 @@ else:
     ids.index = n_elms
     ids.exists = 0"#};
 
-pub fn find_element(
+pub fn search_sorted_optimistic(
     vm: &mut VirtualMachine,
     exec_scopes: &mut ExecutionScopes,
     ids_data: &HashMap<String, HintReference>,
--- a/crates/starknet-os/src/hints/tests.rs
+++ b/crates/starknet-os/src/hints/tests.rs
@@ -296,7 +296,8 @@ pub mod tests {
         let constants = HashMap::new();
 
         let mut rc96_segment = vm.add_memory_segment();
-        rc96_segment.offset = 10;
+        let rc96_segment_size = 10;
+        rc96_segment.offset = rc96_segment_size;
         insert_value_from_var_name(vars::ids::RANGE_CHECK96_PTR, rc96_segment, &mut vm, &ids_data, &ap_tracking)
             .expect("insert_value_from_var_name");
 
@@ -313,7 +314,7 @@ pub mod tests {
         // Make sure existing value isn't overwritten
         assert_eq!(vm.get_maybe(&rc96_base), Some(Felt252::THREE.into()));
 
-        for i in 1..rc96_segment.offset {
+        for i in 1..rc96_segment_size {
             let address = with_offset(rc96_segment, i);
             assert_eq!(vm.get_maybe(&address), Some(Felt252::ZERO.into()));
         }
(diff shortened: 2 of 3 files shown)


crates/starknet-os/src/hints/tests.rs
```diff
@@ -278,4 +282,136 @@ pub mod tests {
...
+        let ap_tracking = ApTracking::default();
+        let constants = HashMap::new();
+
+        let array_ptr = vm.add_memory_segment();
+        insert_value_from_var_name(ARRAY_PTR, array_ptr, &mut vm, &ids_data, &ap_tracking).unwrap();
+        insert_value_from_var_name(ELM_SIZE, Felt252::ONE, &mut vm, &ids_data, &ap_tracking).unwrap();
+        insert_value_from_var_name(N_ELMS, Felt252::THREE, &mut vm, &ids_data, &ap_tracking).unwrap();
+
+        let result = search_sorted_optimistic(&mut vm, &mut exec_scopes, &ids_data, &ap_tracking, &constants);
+        assert!(matches!(result, Err(HintError::AssertionFailed(_))), "{:?}", result);
```
notlesh, line 361: The error message in `AssertionFailed` would be useful to have here for understanding what the test is expecting to fail

crates/starknet-os/src/hints/tests.rs
```diff
@@ -278,4 +282,136 @@ pub mod tests {
...
+        let ids_data = ids_data![ARRAY_PTR, ELM_SIZE];
+        let ap_tracking = ApTracking::default();
+        let constants = HashMap::new();
+
+        let array_ptr = vm.add_memory_segment();
+        insert_value_from_var_name(ARRAY_PTR, array_ptr, &mut vm, &ids_data, &ap_tracking).unwrap();
+        insert_value_from_var_name(ELM_SIZE, Felt252::ZERO, &mut vm, &ids_data, &ap_tracking).unwrap();
+
+        let result = search_sorted_optimistic(&mut vm, &mut exec_scopes, &ids_data, &ap_tracking, &constants);
+        assert!(matches!(result, Err(HintError::AssertionFailed(_))), "{:?}", result);
```
notlesh, line 339: Same as earlier, it would help to replace `_` with the actual error message we expect

crates/starknet-os/src/hints/tests.rs
```diff
@@ -278,4 +282,136 @@ pub mod tests {
...
+        let mut rc96_segment = vm.add_memory_segment();
+        rc96_segment.offset = 10;
+        insert_value_from_var_name(vars::ids::RANGE_CHECK96_PTR, rc96_segment, &mut vm, &ids_data, &ap_tracking)
+            .expect("insert_value_from_var_name");
+
+        let rc96_base = with_offset(rc96_segment, 0);
+        vm.insert_value(rc96_base, Felt252::THREE).expect("insert value at base");
+        for i in 1..rc96_segment.offset {
+            let address = with_offset(rc96_segment, i);
+            assert_eq!(vm.get_maybe(&address), None);
```
notlesh, line 307: In `fill_holes_in_rc96_segment()` we do `vm.insert_value(address, Felt252::ZERO)?;` for each hole, so shouldn't we expect to get `Felt252::ZERO` here instead of `None`?

notlesh: Nevermind, this is before calling the hint :)

crates/starknet-os/src/hints/find_element.rs
```diff
@@ -17,7 +17,7 @@ use crate::utils::get_variable_from_root_exec_scope;
 
 #[rustfmt::skip]
 
-pub const FIND_ELEMENT: &str = indoc! {r#"array_ptr = ids.array_ptr
+pub const SEARCH_SORTED_OPTIMISTIC: &str = indoc! {r#"array_ptr = ids.array_ptr
```
notlesh, line 20: I'm not sure I like the name change here. It better describes what the function does, but it deviates from our strategy of reusing naming conventions and code organization from `cairo-lang` (the fn there is called `find_element` inside `find_element.cairo`).

In other words, if someone is trying to compare our code to that of `cairo-lang` then this new name will not help them.

crates/starknet-os/src/hints/tests.rs
```diff
@@ -278,4 +282,136 @@ pub mod tests {
...
+        vm.add_memory_segment();
+        vm.add_memory_segment();
+
+        let mut exec_scopes = ExecutionScopes::new();
+        let ids_data = ids_data![vars::ids::RANGE_CHECK96_PTR];
+        let ap_tracking = ApTracking::default();
+        let constants = HashMap::new();
+
+        let mut rc96_segment = vm.add_memory_segment();
+        rc96_segment.offset = 10;
```
notlesh, line 299: Minor suggestion: define a constant for 10 and use it throughout the test, I think that is clearer than reusing `rc96_segment.offset` in `for` loops below.

notlesh (changes requested):
Looks good, just a handful of cosmetic suggestions. Feel free to disagree, of course :)

crates/starknet-os/src/hints/find_element.rs
shamsasari: But that's the point I'm making, there are three hints in that file, all doing very similar things, but with just small changes between them. This hint is actually the one in `func search_sorted_optimistic`, hence the name change.

notlesh: Oh, I'm sorry, I misunderstood you earlier. Looks good then!

cairo-lang
shamsasari, line 1: Not sure how this got updated. Let me know if this needs to be reverted.

shamsasari:
PTAL

notlesh (approved):

PR #378, opened 2024-09-18
Fix: find_element(): No value found for key

### LINKED ISSUE #376: bug: find_element(): No value found for key
**Current behavior:**
block_prover is not working for block `164333` and `169206`

**Expected behavior:**
block_prover should be able to process the block

**Steps to reproduce:**
`cargo run --release -p prove_block -- --block-number block_number --rpc-provider rpc_provider_url`

ftheirs:
Problem: Some blocks contain a Declare and in the same block a Deploy transaction. This condition triggers an error in `find_element` hint from cairo-vm

Solution:
The solution consists in setting two hash maps that are used by snos with the correct information.
For declare transactions:
`class_hash_to_compiled_class_hash: HashMap<ClassHash, CompiledClassHash>` needs to be initialized with Zero for all new classes

For deploy transactions:
`compiled_classes: HashMap<CompiledClassHash, GenericCasmContractClass>` needs to be set with the compiled_class_hash from declared class and the corresponding binary.

Issue Number: https://github.com/keep-starknet-strange/snos/issues/376
Closes #376 

## Type

- [ ] feature
- [x] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [ ] yes
- [x] no

### DIFF
--- a/crates/bin/prove_block/src/state_utils.rs
+++ b/crates/bin/prove_block/src/state_utils.rs
@@ -52,7 +52,7 @@ pub(crate) async fn get_formatted_state_update(
         state_diff.declared_classes.iter().map(|declared_item| declared_item.class_hash).collect();
 
     // TODO: Handle deprecated classes
-    let mut class_hash_to_compiled_class_hash: HashMap<Felt252, Felt252> = format_declared_classes(&state_diff);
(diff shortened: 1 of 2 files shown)


HermanObst (approved):
Good work

PR #383, opened 2024-09-24
Test: add pie validity check to prove block

HermanObst:
Problem: We weren't performing validation checks on the PIEs we were generating.
Fix: Adds run_validity_checks() in our prove block tests

### DIFF
--- a/crates/bin/prove_block/tests/prove_block.rs
+++ b/crates/bin/prove_block/tests/prove_block.rs
@@ -47,8 +47,10 @@ use rstest::rstest;
 #[tokio::test(flavor = "multi_thread")]
 async fn test_prove_selected_blocks(#[case] block_number: u64) {
     let endpoint = std::env::var("PATHFINDER_RPC_URL").expect("Missing PATHFINDER_RPC_URL in env");
-    prove_block(block_number, &endpoint, LayoutName::all_cairo)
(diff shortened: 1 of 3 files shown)


ftheirs (approved):
LGTM

notlesh (changes requested):
We should add this to `execute_txs_and_run_os()` so that it will be checked in our integration tests as well

crates/bin/prove_block/tests/prove_block.rs
```diff
@@ -47,8 +47,10 @@ use rstest::rstest;
 #[tokio::test(flavor = "multi_thread")]
 async fn test_prove_selected_blocks(#[case] block_number: u64) {
     let endpoint = std::env::var("PATHFINDER_RPC_URL").expect("Missing PATHFINDER_RPC_URL in env");
-    prove_block(block_number, &endpoint, LayoutName::all_cairo)
+    let (pie, _output) = prove_block(block_number, &endpoint, LayoutName::all_cairo)
         .await
         .map_err(debug_prove_error)
         .expect("Block could not be proven");
+
+    assert!(pie.run_validity_checks().is_ok());
```
notlesh, line 55: Not a big deal, but an `expect()` with a clear message would be better

notlesh (approved):

PR #387, opened 2024-09-27
fix: improve convertion errors and handle invalid ABI

whichqua:
Problem: During convertions between types from blockifier, starknet and cairolang classes, the errors returned are not consistent. This also led to finding an invalid api bug during convertion.

Solution:
- Create better errors for `ConvertionError` and `RpcClientError` and handle them appropriately.
- Default to `None` when abi is not valid JSON.

Issue Number: N/A

## Type

- [ ] feature
- [x] bugfix
- [ ] dev (no functional changes, no API changes)
- [ ] fmt (formatting, renaming)
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [ ] yes
- [x] no

PR #388, opened 2024-09-27
fix: handle contract commitments for contract without data items

whichqua:
Problem: reexecution panics when a contract doesnt have data items

Solution: handle the edge case to prevent panics

Requires #387 and https://github.com/Moonsong-Labs/pathfinder/pull/2

Issue Number: N/A

## Type

- [ ] feature
- [x] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [ ] yes
- [x] no


crates/bin/prove_block/src/reexecute.rs
```diff
@@ -146,28 +146,34 @@ pub(crate) fn format_commitment_facts<H: HashFunctionType>(
...
-            previous_root: self.previous_tree_root,
-            updated_root,
-            tree_height: DEFAULT_STORAGE_TREE_HEIGHT as usize,
-            commitment_facts,
-        })
+            Ok(CommitmentInfo {
+                previous_root: self.previous_tree_root,
+                updated_root,
+                tree_height: DEFAULT_STORAGE_TREE_HEIGHT as usize,
+                commitment_facts,
```
shamsasari, line 167: `If the difference between these two branches is just the `updated_root` and `commitment_facts`, then perhaps return just these two from this if-else expression. That way you can have a single `CommitmentInfo` return.

HermanObst (changes requested):
Please add integration test showing the issue was fixed

PR #391, opened 2024-10-01
fix: `split_commitment` function in KZG data availability implementation

ocdbytes:
Problem:
The bug was that the commitment split returned by this function was wrong. I encountered this when I was integrating the snos with madara orchestrator and performing an e2e test. On the final step of the block processing the contract was giving an error saying that "precompile evaluation failed". After debugging we found out that this is the issue the commitment returned by the snos was wrong.

Issue Number: N/A

## Type

- [ ] feature
- [x] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [ ] yes
- [x] no

### DIFF
--- a/crates/starknet-os/src/hints/kzg.rs
+++ b/crates/starknet-os/src/hints/kzg.rs
@@ -1,5 +1,4 @@
 use std::collections::HashMap;
-use std::io::{self, Read};
 use std::num::ParseIntError;
 use std::path::Path;
 
@@ -16,7 +15,7 @@ use cairo_vm::vm::vm_core::VirtualMachine;
 use cairo_vm::Felt252;
 use indoc::indoc;
 use num_bigint::{BigInt, ParseBigIntError};
-use num_traits::{FromBytes, Num, One, Zero};
+use num_traits::{Num, One, Zero};
 
 use super::vars;
 use crate::execution::helper::ExecutionHelperWrapper;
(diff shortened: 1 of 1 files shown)


ocdbytes:
- The bug was that the commitment split returned by this function was wrong. I encountered this when I was integrating the snos with madara orchestrator and performing an e2e test. On the final step of the block processing the contract was giving an error saying that "precompile evaluation failed". After debugging we found out that this is the issue the commitment returned by the snos was wrong.

- I am not knowledgable enough to get the edge cases for this function as I have very little knowledge of snos but I checked the logic with the e2e testing and even with the contract logic which is mentioned in the comment above the test. Now the split returned by this function is correct and we are able to properly verify the proof on Starknet Core Contract. Please do tell me what edge cases can be here It will also be helpful for me in future and will be an addition to my knowledge.

- I didn't notice any (noticable) performance changes with this fix as it is a simple split function.

If you need any more clarification or there is any confusion we can also discuss this on call. 

Thanks

ocdbytes:
> Thanks for your speedy response.
> 
> > If you need any more clarification or there is any confusion we can also discuss this on call.
> 
> I think you have clarified enough. But I am open to a call, generally just to discuss and understand how you are using snos.

Sure here is my calendly link you can schedule a call whenever you are free: 

https://calendly.com/d/cpt8-mjq-vjf/30min

HermanObst (approved):

HermanObst:
@ocdbytes thanks for this LGTM

notlesh (approved):
Looks good. I'd prefer to use constants as was done previously, but I won't make a big deal out of that. (CC @HermanObst)

HermanObst:
@notlesh agree. 
@ocdbytes please add that and we will merge it :)

Issue #392, opened 2024-10-01
bug: `retdata` is not a Relocatable


whichqua:
## Affected blocks
This issue can be observed in the following blocks:
- 160079
- 160033

## Background
- This issue seems to specifically affect deploy syscalls called in an invoke transaction.
- It occurs in this hint:
```rs
pub const CHECK_NEW_DEPLOY_RESPONSE: &str = indoc! {r#"
	# Check that the actual return value matches the expected one.
	expected = memory.get_range(
	    addr=ids.response.constructor_retdata_start,
	    size=ids.response.constructor_retdata_end - ids.response.constructor_retdata_start,
	)
	actual = memory.get_range(addr=ids.retdata, size=ids.retdata_size)
	assert expected == actual, f'Return value mismatch; expected={expected}, actual={actual}.'"#
};
```
- *The issue seems to `constructor_args` being an empty vector.*

## Next steps

- [x] Write integration tests covering deploy_contact syscall in an invoke transaction.
- [x] Attempt to do a deploy with empty args?

PR #393, opened 2024-10-02
fix: Identifier `retdata` is not relocatable

whichqua:
Problem: 
This issue occurs with the following conditions
1. `deploy_contract` is called inside an invoke tx.
2. The contract being deployed has empty calldata array.

Solution:
- [x] Replicate the issue as an integration test via this commit [test: add tests for contract with no calldata](https://github.com/keep-starknet-strange/snos/commit/1b4401ea3b0253feb3034266c84a6ca960a3f742)
- [ ] Handle this edge case.

### Rc::new(os_input)

Note that this PR also wraps the os_input in a Rc. This was necessary(-ish) so that the deploy syscall could inspect the class info, but *it should also have a big improvement to performance.* See 3ddd5ef (and 8562944).

Issue Number: #392 

## Type

- [ ] feature
- [x] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [ ] yes
- [ ] no


notlesh (approved):

crates/starknet-os/src/hints/execution.rs
```diff
@@ -1119,6 +1121,21 @@ pub fn check_new_deploy_response(
...
+    // points. Immediately following this, `add_relocation_rule()` will be called with src=0, dest=0
+    // and it will also no-op.
+    let retdata_start_key: Relocatable =
+        (response_ptr + new_syscalls::DeployResponse::constructor_retdata_start_offset())?;
+    let maybe_retdata_start = vm.get_maybe(&retdata_start_key).unwrap();
+    let zero = MaybeRelocatable::Int(Felt252::ZERO);
+    if maybe_retdata_start == zero {
+        log::warn!("retdata_start is 0, not a relocatable. doing nothing");
+        return Ok(());
+    }
```
HermanObst, line 1137: @notlesh could you add more context to this? 
It feels weird that in a check function, we got something that we don not expect and we do nothing.

Ideally you add a link to the code that explains why this "error bypass" is correct

crates/starknet-os/src/hints/execution.rs
```diff
@@ -1119,6 +1121,21 @@ pub fn check_new_deploy_response(
     _constants: &HashMap<String, Felt252>,
 ) -> Result<(), HintError> {
     let response_ptr = get_ptr_from_var_name(vars::ids::RESPONSE, vm, ids_data, ap_tracking)?;
```
HermanObst, line 1123: @notlesh cant we use the "need_retdata_hack" here?

crates/starknet-os/src/execution/syscall_handler.rs
```diff
@@ -291,12 +292,56 @@ where
...
+
+        // Bugfix:
+        // When retdata size is 0, the OS will return retdata as a Int(0) instead of a relocatable
+        // as a result of `cast(0, felt*)`. To work around this, we can write the same here so that
+        // later the OS will assert this value is the same as retdata; that is, both need to be the
+        // same value which is Int(0).
+        if response.need_retdata_hack {
+            log::warn!("Writing Felt::ZERO instead of pointer since retdata size is 0");
+            write_felt(vm, ptr, Felt252::ZERO)?; // casted pointer
+            write_felt(vm, ptr, Felt252::ZERO)?; // size
```
HermanObst, line 341: @notlesh could you add link to the OS code that returns a NULL pointer?

HermanObst (changes requested):

crates/starknet-os/src/execution/helper.rs
```diff
@@ -28,6 +29,7 @@ where
     PCS: PerContractStorage,
 {
     pub _prev_block_context: Option<BlockContext>,
+    pub os_input: Option<Rc<StarknetOsInput>>,
```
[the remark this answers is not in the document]
notlesh: That avoids cloning it, which is the part of this PR that improves performance

crates/starknet-os/src/execution/syscall_handler.rs
```diff
@@ -291,12 +292,56 @@ where
...
+        } else {
+            write_segment(vm, ptr, response.constructor_retdata)?;
+        }
```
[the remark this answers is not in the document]
notlesh: Which part? The syscall is specific to the OS.

If you mean the underlying problem (which is that `cast(0, felt*)` is represented by `cairo-vm` as an `Int(0)` then yes, I think that could be an upstream bug, but I'm not sure. It's worth asking about...

crates/starknet-os/src/hints/execution.rs
notlesh: I don't think so. The `need_retdata_hack` is part of the syscall struct, which has disappeared by this point. It's possible that we could write something to the vm to use later, but I think that's a bad idea.

crates/starknet-os/src/execution/syscall_handler.rs
```diff
@@ -291,12 +292,56 @@ where
             "No more deployed contracts available to replay".to_string().into_boxed_str(),
         ))?;
 
-        Ok(DeployResponse { contract_address, constructor_retdata })
+        // TODO: hideous clone
+        let mut need_retdata_hack = false;
+        execution_helper.os_input.clone().inspect(|os_input| {
```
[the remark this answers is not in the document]
notlesh: Fixed in [97b7f14](https://github.com/keep-starknet-strange/snos/pull/393/commits/97b7f14ed0e85d3e8634ac527c97f564c031bbaa)

crates/starknet-os/src/execution/helper.rs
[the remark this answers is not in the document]
notlesh: To elaborate: the OsInput is now used in both `ExecutionHelper` and in `run_os()` so this avoids having to have two copies of it in those cases

crates/starknet-os/src/hints/execution.rs
notlesh: Yeah, it is a weird place for part of this hack. I elaborated, see [3ee3f5d](https://github.com/keep-starknet-strange/snos/pull/393/commits/3ee3f5df5e217a281b9cfd1572759e44a6a92b77)

crates/starknet-os/src/execution/syscall_handler.rs
notlesh: Added link here: [845bcfa](https://github.com/keep-starknet-strange/snos/pull/393/commits/845bcfa7cd5b0a26e0e65721f2e9027bbec2d648)

HermanObst (approved):

PR #401, opened 2024-10-09
fix: timestamp and block number mismatch

HermanObst:
Problem: Some blocks were failing because the get_block_number and get_block_timestamp were retrieving different values than the SNOS was expecting.
This is related to the fact that two modes of execution exist: Execute (were the expected behaviour is to get the actual block number and timestamp) and Validate (were the expectation is to get an rounding of them)
 
Fix: modify the given syscalls to fetch the blockInfo from the ExecutionInfo struct from SNOS


crates/starknet-os/src/execution/deprecated_syscall_handler.rs
```diff
@@ -136,9 +137,11 @@ where
         let syscall_handler = self.deprecated_syscall_handler.read().await;
 
         let block_number = syscall_handler.block_info.block_number;
+        let rounded_block_number = (block_number.0 / VALIDATE_BLOCK_NUMBER_ROUNDING ) * VALIDATE_BLOCK_NUMBER_ROUNDING;
```
shamsasari, line 140: Can you use `prev_multiple_of` instead?

crates/starknet-os/src/execution/deprecated_syscall_handler.rs
```diff
@@ -15,6 +15,7 @@ use crate::cairo_types::syscalls::{
     TxInfo,
 };
 use crate::starknet::starknet_storage::PerContractStorage;
+use crate::execution::constants::{VALIDATE_BLOCK_NUMBER_ROUNDING, VALIDATE_TIMESTAMP_ROUNDING};
```
notlesh, line 18: We already discussed this, but these should be able to come from the constants in `cairo-lang`. We should do that if at all possible, but if not we should leave a comment about that here.

crates/starknet-os/src/execution/deprecated_syscall_handler.rs
```diff
@@ -134,11 +135,16 @@ where
 
     pub async fn get_block_number(&self, syscall_ptr: Relocatable, vm: &mut VirtualMachine) -> Result<(), HintError> {
         let syscall_handler = self.deprecated_syscall_handler.read().await;
+        let execution_helper = syscall_handler.exec_wrapper.execution_helper.read().await;
 
-        let block_number = syscall_handler.block_info.block_number;
+        let execution_info_ptr = execution_helper
+            .call_execution_info_ptr
+            .ok_or(HintError::SyscallError("Execution info pointer not set".to_string().into_boxed_str()))?;
+        let block_info_pointer = vm.get_relocatable((execution_info_ptr + ExecutionInfo::block_info_offset())?)?;
```
notlesh, line 143: Super nit: we usually `_ptr` instead of `_pointer`

crates/starknet-os/src/execution/deprecated_syscall_handler.rs
```diff
@@ -149,12 +155,18 @@ where
...
         let syscall_handler = self.deprecated_syscall_handler.read().await;
+        let execution_helper = syscall_handler.exec_wrapper.execution_helper.read().await;
 
-        let block_timestamp = syscall_handler.block_info.block_timestamp;
+        let execution_info_ptr = execution_helper
+            .call_execution_info_ptr
+            .ok_or(HintError::SyscallError("Execution info pointer not set".to_string().into_boxed_str()))?;
+        let block_info_pointer = vm.get_relocatable((execution_info_ptr + ExecutionInfo::block_info_offset())?)?;
+        let block_timestamp =
+            vm.get_integer((block_info_pointer + BlockInfoStruct::block_timestamp_offset())?)?.into_owned();
```
notlesh, line 165:
```suggestion
        let block_info_ptr = vm.get_relocatable((execution_info_ptr + ExecutionInfo::block_info_offset())?)?;
        let block_timestamp =
            vm.get_integer((block_info_ptr + BlockInfoStruct::block_timestamp_offset())?)?.into_owned();
```

tests/integration/deprecated_syscalls_tests.rs
```diff
@@ -95,14 +95,14 @@ async fn test_syscall_get_block_number_cairo0(
     let sender_address = initial_state.deployed_cairo0_contracts.get("account_with_dummy_validate").unwrap().address;
     let contract_address = initial_state.deployed_cairo0_contracts.get("test_contract").unwrap().address;
 
-    let expected_block_number = felt!(block_context.block_info().block_number.0);
+    let block_number = block_context.block_info().block_number.0;
 
-    let tx_version = TransactionVersion::ZERO;
+    let tx_version = TransactionVersion::THREE;
```
notlesh, line 100: Is this necessary?

HermanObst:
Thanks @notlesh and @whichqua. Yes, the tx version = 3 was something I change while exploring possible cases and then I forget.  fixed :)

PR #402, opened 2024-10-10
Add cairo0_run binary

notlesh:
This PR adds a `cairo0_run` binary which runs a compiled cairo0 program directly against `cairo-vm`. It does use the SNOS hint processor, which allows interaction with our hints (this was useful to pause in our `breakpont()` hint to inspect the VM).

This probably isn't specific to `cairo0`, but I found that the `cairo1_run` in `cairo-vm` did not work for cairo0...

## Type

- [ ] feature
- [ ] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [ ] yes
- [x] no


crates/bin/cairo0_run/src/main.rs
```diff
@@ -0,0 +1,87 @@
...
+        cairo_run_config.proof_mode,
+        cairo_run_config.trace_enabled,
+    )
+    .map_err(|e| SnOsError::Runner(e.into()))?;
+
+    // Init the Cairo VM
+    let end = cairo_runner.initialize(allow_missing_builtins).map_err(|e| SnOsError::Runner(e.into()))?;
+
+    // Run the Cairo VM
+    let mut hint_processor = hints::SnosHintProcessor::<DummyPCS>::default();
```
whichqua, line 57: Why are we using a Dummy storage?

notlesh: Because this isn't meant to do any storage ops, it's really just meant to run cairo0 programs. I suppose some OS features could be supported, but that should be implemented when needed...

HermanObst (approved):

GMKrieger (approved):
LGTM

PR #406, opened 2024-10-15
Moved compiled snos import to allowing using snos as a dependency

matzayonc:
Issue Number: N/A

## Type

- [ ] feature
- [x] bugfix
- [] dev (no functional changes, no API changes)
- [] fmt (formatting, renaming)
- [ ] build
- [ ] docs
- [ ] testing

## Description
I moved the `include_bytes!()` with `snos` import to tests and binary, because it is not present when used as a dependency.

## Breaking changes?

- [x] yes
- [ ] no


matzayonc:
Moved to #407

PR #409, opened 2024-10-15
fix: case when both `l1_gas` and `l1_gas_price` are both 0 for l1_handler tx

whichqua:
## Problem:
This PR addresses the scenario where both `l1_gas` and `l1_data_gas` are zero during fee calculation for L1 handler transactions. In such cases, multiplying the values would result in a fee of zero, which causes the transaction to fail during the blockifier execution as seen [here](https://github.com/starkware-libs/sequencer/blob/b5a877719dc2ce5b1ca833f14d9473c1f1c27059/crates/blockifier/src/transaction/transaction_execution.rs#L166) and [here](https://github.com/eqlabs/pathfinder/blob/eb81bf149fe516c3542a90a5c1715c5a3a141d0b/crates/rpc/src/executor.rs#L548).

## Solution:
- Added handling for the case where both `l1_gas` and `l1_data_gas` are zero, assigning a default fee of `1_000_000_000_000u128` to prevent failures.
- The following match arms have been adjusted to ensure correct fee calculation:
  - `(0, 0) => 1_000_000_000_000u128`
  - `(0, l1_data_gas)` calculates fee using `eth_l1_data_gas_price`.
  - `(l1_gas, 0)` calculates fee using `eth_l1_gas_price`.

Issue Number: N/A

## Type

- [ ] feature
- [x] bugfix
- [ ] build
- [ ] docs
- [x] testing

## Breaking changes?

- [ ] yes
- [x] no

### DIFF
--- a/crates/rpc-replay/src/transactions.rs
+++ b/crates/rpc-replay/src/transactions.rs
@@ -244,6 +244,7 @@ fn l1_handler_to_blockifier(
         // Having the fee as 0 for L1 handler will fail on the blockifier execution
         // Learn more:
         // https://github.com/starkware-libs/sequencer/blob/b5a877719dc2ce5b1ca833f14d9473c1f1c27059/crates/blockifier/src/transaction/transaction_execution.rs#L166
+        // https://github.com/eqlabs/pathfinder/blob/eb81bf149fe516c3542a90a5c1715c5a3a141d0b/crates/rpc/src/executor.rs#L548
         // The comment(which is not very helpful) on the line above is:
         // // For now, assert only that any amount of fee was paid.
         // More investigations are recommended


notlesh (approved):

ftheirs:
This looks good! Let's merge the latest changes from main 💪

crates/rpc-replay/src/transactions.rs
```diff
@@ -238,12 +238,19 @@ fn l1_handler_to_blockifier(
...
+    let fee = match (l1_gas, l1_data_gas) {
+        // There are the cases where both these values are zero and that means no matter what we multiply,
+        // we will get a value of 0.
+        // Having the fee as 0 for L1 handler will fail on the blockifier execution
+        // Learn more:
+        // https://github.com/starkware-libs/sequencer/blob/b5a877719dc2ce5b1ca833f14d9473c1f1c27059/crates/blockifier/src/transaction/transaction_execution.rs#L166
+        // The comment(which is not very helpful) on the line above is:
+        // // For now, assert only that any amount of fee was paid.
+        // More investigations are recommended
+        (0, 0) => 1u128,
```
HermanObst, line 250: @whichqua what happen if we return `0u128`?

ftheirs: There will be an error in blockifier:
https://github.com/starkware-libs/sequencer/blob/b5a877719dc2ce5b1ca833f14d9473c1f1c27059/crates/blockifier/src/transaction/transaction_execution.rs#L166

notlesh: It would be good to include this link in the explanation of the workaround

ftheirs (approved):

HermanObst (approved):
LGTM

PR #411, opened 2024-10-17
handle underflow while requesting older block

HermanObst:
Problem: The OS require the hash and block number of a older block (current - BUFFER) to perform some validations.
If the Block that we want to prove execution is lower that BUFFER, the OS correctly detects this and do nothing, but our prove_block binary was not having this into account.

Fix: Check if current_block < BUFFER and default to 1 in this case.

### DIFF
--- a/crates/bin/prove_block/src/lib.rs
+++ b/crates/bin/prove_block/src/lib.rs
@@ -131,9 +131,17 @@ pub async fn prove_block(
     };
 
     // We only need to get the older block number and hash. No need to fetch all the txs
+    // This is a workaorund to catch the case where the block number is less than the buffer and still preserve the check
+    // The OS will also handle the case where the block number is less than the buffer.
+    let older_block_number = if block_number <= STORED_BLOCK_HASH_BUFFER {
+        1
+    } else {
+        block_number - STORED_BLOCK_HASH_BUFFER
+    };
+
     let older_block = match rpc_client
         .starknet_rpc()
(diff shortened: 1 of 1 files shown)


crates/bin/prove_block/src/lib.rs
```diff
@@ -131,21 +131,21 @@ pub async fn prove_block(
...
-            panic!("Block is still pending!");
-        }
-    };
-
-    let block_context = build_block_context(chain_id.clone(), &block_with_txs, starknet_version);
-
+    // This is a workaorund to catch the case where the block number is less than the buffer and still preserve the check
+    // The OS will also handle the case where the block number is less than the buffer.
+    let older_block_number =
+        if block_number <= STORED_BLOCK_HASH_BUFFER { 1 } else { block_number - STORED_BLOCK_HASH_BUFFER };
```
notlesh, line 137: This is what the OS does for a similar check:

```
    tempvar old_block_number = block_context.block_info_for_execute.block_number -
        STORED_BLOCK_HASH_BUFFER;
    let is_old_block_number_non_negative = is_nn(old_block_number);
    if (is_old_block_number_non_negative == FALSE) {
        // Not enough blocks in the system - nothing to write.
        return ();
    }
```

What we do in `prove_block` with this is feed it to `ExecutionHelper`, which is used later by the OS just after the code snippet above to set the block number/hash. Since it's protected by the check above, I think it should be ok.

However, it may be cleaner to make `old_block_number` and `old_block_hash` Optional, and it would also be good to avoid doing an RPC query in this case.

whichqua (approved):
LGTM

notlesh (approved):

PR #414, opened 2024-10-28
Test: compare output segment binary

HermanObst:
Problem: We need to check that the output segment of our pies are the same than the ones generates by starkware

Solution: Add output_segment binary that is in change of taking a pie zip, gets it block number, run SNOS and compare the output segment


notlesh (changes requested):
We should avoid merging these large binary files into `main`, even if we intend to remove them later. Doing so will permanently bloat the git repository (even after they are removed).

HermanObst:
@notlesh done!

compare_output.sh
```diff
@@ -0,0 +1,65 @@
...
+lcd $LOCAL_DIR
+get $file
+EOF
+
+    # Check if download succeeded
+    if [ $? -eq 0 ]; then
+        echo "Successfully downloaded $file" 
+
+        # Run the Rust program with the downloaded file using the specified command
+        if RUST_LOG="debug,minilp::solver=info" cargo run -p output_segment --release -- --rpc-provider "http://81.16.176.130:9545" --pie-path "$LOCAL_DIR/$file"; then
```
notlesh, line 48: This IP address should be passed in by the caller :)

compare_output.sh
```diff
@@ -0,0 +1,65 @@
+#!/bin/bash
+
```
notlesh, line 2: Some documentation could be useful here (how to use the script, what it does), or consider a `--help` output.

compare_output.sh
```diff
@@ -0,0 +1,65 @@
...
+# Create local directory if it does not exist
+mkdir -p "$LOCAL_DIR"
+
+# Start the SFTP session to list files
+# echo "Starting SFTP session to list files..." | tee -a "$LOG_FILE"
+# sftp $SFTP_USER@$SFTP_SERVER << EOF > temp_file_list.txt
+# cd $REMOTE_PATH
+# ls
+# EOF
```
notlesh, line 30: Eww, it looks like sftp doesn't make it easy to list directory contents. Maybe an alternative is available (`scp` or `ssh`) where you could do this as part of the script itself?

crates/bin/output_segment/src/main.rs
```diff
@@ -0,0 +1,82 @@
...
+#[tokio::main]
+async fn main() {
+    init_logging();
+
+    let args = Args::parse();
+
+    let zip_pie_path = args.zip_pie_path;
+    let endpoint = args.rpc_provider;
+
+    let reference_pie_bytes = fs::read(zip_pie_path).unwrap();
```
notlesh, line 39: Some ideas instead of unwrap:

* `expect()`
* propagate error (make `main()` return some type of `Result`)
* use something like `anyhow`: https://docs.rs/anyhow/latest/anyhow/

crates/bin/output_segment/src/main.rs
```diff
@@ -0,0 +1,82 @@
...
+
+    let (snos_pie, _snos_output) =
+        prove_block(DEFAULT_COMPILED_OS, block_number, &endpoint, LayoutName::all_cairo, true)
+            .await
+            .map_err(debug_prove_error)
+            .expect("OS generate Cairo PIE");
+
+    snos_pie.run_validity_checks().expect("Valid SNOS PIE");
+
+    let output_segment_index = 2;
```
notlesh, line 56: Is there any reference you can add about this?

crates/bin/output_segment/src/main.rs
```diff
@@ -0,0 +1,82 @@
...
+
+    assert_eq!(
+        get_memory_segment(&reference_pie, output_segment_index),
+        get_memory_segment(&snos_pie, output_segment_index)
+    );
+
+    log::info!("SNOS Pie has the same output as reference pie");
```
notlesh, line 63: This totally needs a checkmark emoji!

run_output_check.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+
```
notlesh, line 2: Same as the other script, a short description would go a long way

notlesh (commented):
Looks good, I left a bunch of small nits but nothing critical

notlesh (approved):

crates/bin/output_segment/src/main.rs
HermanObst: You wanted, you have it 😎

crates/bin/prove_block/tests/prove_block.rs
```diff
@@ -1,7 +1,6 @@
 use cairo_vm::types::layout_name::LayoutName;
-use cairo_vm::types::relocatable::MaybeRelocatable;
 use cairo_vm::vm::runners::cairo_pie::CairoPie;
-use prove_block::{debug_prove_error, prove_block};
+use prove_block::{debug_prove_error, get_memory_segment, prove_block};
 use rstest::rstest;
 
 pub const DEFAULT_COMPILED_OS: &[u8] = include_bytes!("../../../../build/os_latest.json");
```
ftheirs, line 6: super picky but.. since you've moved `DEFAULT_COMPILED_OS`, you can remove this constant and use the one that is in lib.rs

ftheirs (approved):

crates/bin/prove_block/tests/prove_block.rs
HermanObst: Totally right. Good catch :)

PR #415, opened 2024-10-28
Fix: Serialize pie function

HermanObst:
Problem: serialize/deserialize is not properly working

Fix: cairo-vm serialize function was not expecting new builtins (fixed [here](https://github.com/Moonsong-Labs/cairo-vm/pull/48/files))

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,7 +35,7 @@ cairo-lang-starknet-classes = { version = "=2.8.2" }
 cairo-lang-utils = { version = "=2.8.2" }
 cairo-lang-casm = { version = "=2.8.2" }
 cairo-type-derive = { version = "0.1.0", path = "crates/cairo-type-derive" }
-cairo-vm = { git = "https://github.com/Moonsong-Labs/cairo-vm", branch = "notlesh/segment-arena-relocation-fix", features = ["cairo-1-hints", "extensive_hints", "mod_builtin"] }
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -35,7 +35,7 @@ cairo-lang-starknet-classes = { version = "=2.8.2" }
 cairo-lang-utils = { version = "=2.8.2" }
 cairo-lang-casm = { version = "=2.8.2" }
 cairo-type-derive = { version = "0.1.0", path = "crates/cairo-type-derive" }
-cairo-vm = { git = "https://github.com/Moonsong-Labs/cairo-vm", branch = "notlesh/segment-arena-relocation-fix", features = ["cairo-1-hints", "extensive_hints", "mod_builtin"] }
+cairo-vm = { git = "https://github.com/Moonsong-Labs/cairo-vm", branch = "herman/fix-pie-serialization", features = ["cairo-1-hints", "extensive_hints", "mod_builtin"] }
```
ftheirs, line 38: This is good but maybe we can point to a dev or a more stable branch (at least a better name)

ftheirs (approved):

notlesh (approved):

Cargo.toml
shamsasari: We really should be using a default "snos-main" branch now.

Issue #405, opened 2024-10-14
bug: enable 0 gas price blocks


apoorvsadana:
**SNOS version:**

3bd95bfb315a596519b5b1d3c81114fbfa06f8a6

**Current behavior:**

Running a block with gas_price 0 panics over here as it expects the price to be > 1.

**Expected behavior:**

It's possible to have a 0 gas price if 

1. The chain is just starting up. The fee token isn't deployed at this point and you need to setup using 0 gas fees. For example, check first tx of Sepolia [here](https://sepolia.starkscan.co/tx/0x0656e113cb27707d2147c271a79c51d1069b0273ae447b965e15154a17b3ec01).
2. The app chain just wants to run without any fees
3. We don't have a custom fee token like `strk` so those prices are 0 (will need to check how do V3 txs work here).

**Steps to reproduce:**

It's a little hacky to do it right now, at least the way I did it

1. Run Madara in devnet
2. Connect pathfinder to this Madara
3. Create blocks with 0 fees
4. Run SNOS

However, once the <10 block issue is fixed maybe we can run it on blocks from 0-5 on Sepolia, it will be easier.

**Related code:**

https://github.com/keep-starknet-strange/snos/blob/3bd95bfb315a596519b5b1d3c81114fbfa06f8a6/crates/rpc-replay/src/block_context.rs#L29

PR #416, opened 2024-10-29
allow zero as gas prices

ftheirs:
Problem: Gas prices can potentially be zero under certain conditions. See detailed explanation [here](https://github.com/keep-starknet-strange/snos/issues/405).

Solution: Since the GasPrice struct uses NonZeroU128 types for its inner elements, we need to use a default fallback value in these cases, similar to the approach taken in [Papyrus](https://github.com/starkware-libs/sequencer/blob/7218aa1f7ca3fe21c0a2bede2570820939ffe069/crates/papyrus_execution/src/lib.rs#L363-L371).

This can be reproduced on Sepolia for blocks 0~5 but the Starknet version from those blocks (0.12.x) is not supported by SNOS yet.

Issue Number: https://github.com/keep-starknet-strange/snos/issues/405
Closes https://github.com/keep-starknet-strange/snos/issues/405

## Type

- [ ] feature
- [x] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [ ] yes
- [x] no


crates/rpc-replay/src/block_context.rs
```diff
@@ -25,11 +27,17 @@ pub fn build_block_context(
...
-            eth_l1_data_gas_price: felt_to_u128(&block.l1_data_gas_price.price_in_wei).try_into().unwrap(),
-            strk_l1_data_gas_price: felt_to_u128(&block.l1_data_gas_price.price_in_fri).try_into().unwrap(),
+            eth_l1_gas_price: felt_to_u128(&block.l1_gas_price.price_in_wei).try_into().unwrap_or(NonZeroU128::MIN),
+            strk_l1_gas_price: felt_to_u128(&block.l1_gas_price.price_in_fri).try_into().unwrap_or(NonZeroU128::MIN),
+            eth_l1_data_gas_price: felt_to_u128(&block.l1_data_gas_price.price_in_wei)
+                .try_into()
+                .unwrap_or(NonZeroU128::MIN),
+            strk_l1_data_gas_price: felt_to_u128(&block.l1_data_gas_price.price_in_fri)
+                .try_into()
+                .unwrap_or(NonZeroU128::MIN),
```
notlesh, line 40: I think we should be more careful with these checks because there are two cases that we are treating the same when we use `try_into().unwrap_or(NonZeroU128::MIN)`:

1. the intended case: `try_into()` errors because of the value `1`
2. the unintended case: `try_into()` errors because the felt overflows (is `> u128::MAX`)

The latter case should be handled more gracefully.

ftheirs: Good catch 😉 I'll fix it asap

crates/rpc-replay/src/block_context.rs
```diff
@@ -1,19 +1,32 @@
...
+
+fn felt_to_gas_price(price: &Felt) -> Result<NonZeroU128, FeltConversionError> {
+    // Inspiration taken from Papyrus:
+    // https://github.com/starkware-libs/sequencer/blob/7218aa1f7ca3fe21c0a2bede2570820939ffe069/crates/papyrus_execution/src/lib.rs#L363-L371
+    if *price == Felt::ZERO {
+        return Ok(NonZeroU128::MIN);
+    }
+
+    // If felt_to_u128 conversion is ok, it won't fail cause we're catching the zero above
+    Ok(NonZeroU128::new(felt_to_u128(price)?).unwrap())
```
HermanObst, line 22: Why don't handle the error gracefully?

crates/rpc-replay/src/block_context.rs
```diff
@@ -50,5 +63,46 @@ pub fn build_block_context(
...
+
+        // Call this function must not fail
+        let block_context = build_block_context(chain_id, &block, starknet_version).unwrap();
+
+        // Verify that gas prices were set to NonZeroU128::MIN
+        assert_eq!(block_context.block_info().gas_prices.eth_l1_gas_price, NonZeroU128::MIN);
+        assert_eq!(block_context.block_info().gas_prices.strk_l1_gas_price, NonZeroU128::MIN);
+        assert_eq!(block_context.block_info().gas_prices.eth_l1_data_gas_price, NonZeroU128::MIN);
+        assert_eq!(block_context.block_info().gas_prices.strk_l1_data_gas_price, NonZeroU128::MIN);
+    }
```
HermanObst, line 107: Cool! Could we add a test that checks that, if we have correct values, the block context is updated accordingly?

crates/rpc-replay/tests/test_replay_block.rs
```diff
@@ -28,7 +28,7 @@ async fn test_replay_block() {
     let state_reader = AsyncRpcStateReader::new(rpc_client.clone(), previous_block_id);
     let mut state = CachedState::from(state_reader);
 
-    let block_context = build_block_context(ChainId::Sepolia, &block_with_txs, StarknetVersion::V0_13_1);
+    let block_context = build_block_context(ChainId::Sepolia, &block_with_txs, StarknetVersion::V0_13_1).unwrap();
```
HermanObst, line 31: Handle error

HermanObst (changes requested):
Looking good! Left some request for changes

HermanObst (approved):

notlesh (approved):
This seems good for now, but the rules around zero gas are not that clear to me. If gas can actually be zero and we need to support that, then it seems the upstream `GasPrices` struct should be changed to reflect this.

PR #417, opened 2024-10-30
Read PIE from sftp

GMKrieger:
Issue Number: N/A

## Type

- [x] feature
- [ ] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Description

This PR adds some new code that encapsulates Herman's work on comparing two PIE's. It works like this:

There's a list of all PIE's present on the remote starkware sftp server. It goes line by line and concatenates the file name on the path provided when running, downloads the file (sftp doesn't allow streaming) and compares it to the snos PIE. The file is deleted at the end.

The logs are printed to a file name `all_logs.log` on `result-logs` subfolder. When the code has completed a cycle, it will write the logs for this execution to another file, named either `SUCCESS` or `FAILURE` followed by the pie name.

Currently, the paths are hard coded, so the `cargo run` command must be run on the root of the repo.

## Breaking changes?

- [ ] yes
- [ ] no


notlesh (commented):
Kudos for getting sftp to work in rust :)

This seems like a step backwards in flexibility, though. A PIE doesn't have to come from sftp, so this change currently forces the user to make it available via sftp.

The Unix philosophy seems relevant here: do one thing and do it well. SNOS doesn't need to know anything about sftp in order to do its job, instead it could operate on something generic like either STDIN (pipe the PIE file) or a file (read from some file system). If sftp is required, then the user can easily handle this detail on the CLI / bash script / etc.

That said, if the sftp part turns out to be best done in rust, I think it would be fine to do so, but I would make that a separate feature/crate to keep the complexity isolated.

HermanObst:
@GMKrieger closed in favor of #414 
As @notlesh a bash script was a better solution. Anyway, awesome job with this PR. Lets keep it up!!

PR #419, opened 2024-11-04
Remove retdata hacks

notlesh:
This is a "preview" of fixes for the retdata issue [from upstream](https://github.com/lambdaclass/cairo-vm/pull/1862).

It undoes a lot of the workarounds (affectionately known as the "retdata hacks") from #393 which aren't needed with the upstream fix mentioned above.

## TODO
 - [ ] Bump `cairo-vm` so that it includes the above-mentioned fix


HermanObst:
@notlesh I think we do not need os_input as part of executionHelper anymore

HermanObst (approved):
LGTM! Still we need to revert the changes that introduce the os_input as part of execution_helper (and that also wrap it into an Arc)

This can be done in a different PR

ftheirs (approved):

PR #421, opened 2024-11-04
Add os hints tests

GMKrieger:
Issue Number: N/A

## Type

- [ ] feature
- [ ] bugfix
- [ ] build
- [ ] docs
- [x] testing

## Description

Adds new unit tests to `crates/starknet-os/hints/os.rs`

## Breaking changes?

- [ ] yes
- [x] no


crates/starknet-os/src/hints/os.rs
```diff
@@ -84,3 +84,135 @@ pub fn set_ap_to_new_block_hash(
...
+    use cairo_vm::vm::vm_core::VirtualMachine;
+    use cairo_vm::Felt252;
+    use rstest::{fixture, rstest};
+
+    use super::*;
+    use crate::utils::get_variable_from_root_exec_scope;
+
+    #[fixture]
+    fn os_input() -> StarknetOsInput {
+        StarknetOsInput {
```
ftheirs, line 106: I think that StarknetOsInput doesn't derive from Default but that might be handy for other tests. Besides, it will look cleaner if the variable is initialized only with the needed values

ftheirs:
It looks great 💪

crates/starknet-os/src/hints/os.rs
notlesh: Note that the fixture here is different than what `#[derive(Default)]` would be (`new_block_hash`, `prev_block_hash`, `full_output` are not defaults)

GMKrieger: I don't understand your request. The struct needs to be initialized with all fields and has no new method to return an empty struct that I can fill.

ftheirs (approved):

PR #426, opened 2024-11-06
Local devnet: fix block less than 10

ftheirs:
Until block 10,  contract_data field  from storage_proof delivered by Pathfinder is None. Handle this situation with a default value.

Issue Number: N/A

## Type

- [ ] feature
- [x] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [ ] yes
- [x] no


crates/bin/prove_block/src/reexecute.rs
```diff
@@ -147,8 +147,11 @@ pub(crate) fn format_commitment_facts<H: HashFunctionType>(
 impl PerContractStorage for ProverPerContractStorage {
     async fn compute_commitment(&mut self) -> Result<CommitmentInfo, CommitmentInfoError> {
         // TODO: error code
-        let contract_data =
-            self.storage_proof.contract_data.as_ref().expect("storage proof should have a contract_data field");
+        let contract_data = match self.storage_proof.contract_data.as_ref() {
+            None => &ContractData::default(),
+            Some(data) => data,
+        };
```
shamsasari, line 153: Do `self.storage_proof.contract_data.as_ref().unwrap_or_default()` instead

crates/bin/prove_block/src/reexecute.rs
```diff
@@ -147,8 +147,11 @@ pub(crate) fn format_commitment_facts<H: HashFunctionType>(
 impl PerContractStorage for ProverPerContractStorage {
     async fn compute_commitment(&mut self) -> Result<CommitmentInfo, CommitmentInfoError> {
         // TODO: error code
```
HermanObst, line 149: @ftheirs please add a comment that explain the motivation of this fix. 
In general storage_proofs **needs** to have this field.

HermanObst: Specifically which contracts are and why the start having contract_data in block 10

crates/bin/prove_block/src/reexecute.rs
ftheirs: Since we are working with `&ContractData`, it is not straightforward to use methods like `unwrap_or_default` or `unwrap_or`. The variable `contract_data` is expected to be a reference, and implementing `Default` for `&ContractData` directly is not possible because it is a reference, not an owned type. While there are some workarounds to simulate this behavior, the primary goal should be to keep the code simple and readable.

crates/bin/prove_block/src/reexecute.rs
ftheirs: In pathfinder, the structure from the requested type is defined with Opt<ContractData>:
https://github.com/Moonsong-Labs/pathfinder/blob/c45b23d2d6407079ad19f84ef0b23f913063d7cd/crates/rpc/src/pathfinder/methods/get_proof.rs#L182-L199
Besides, when we work with a devnet, we're returning in an early part of the function. We get a None when we retrieve the `contract_state_hash` from the DB, and this leads to the early return where most of the fields are empty/none
https://github.com/Moonsong-Labs/pathfinder/blob/c45b23d2d6407079ad19f84ef0b23f913063d7cd/crates/rpc/src/pathfinder/methods/get_proof.rs#L283-L294

When we bump into this condition, the fix just create a default ContractData structure with all fields in zero

whichqua (approved):
LGTM!

HermanObst (approved):
LGTM

PR #427, opened 2024-11-07
No deploy nullptrs

notlesh:
This PR removes the retdata hacks (as done in #419) but also makes a small change to `cairo-lang` to avoid `cast(0, felt*)` during deploys.

The changes to the OS are effectively:

```diff
diff --git a/src/starkware/starknet/core/os/execution/deprecated_execute_entry_point.cairo b/src/starkware/starknet/core/os/execution/deprecated_execute_entry_point.cairo
index d147c23f7b..e329f5855d 100644
--- a/src/starkware/starknet/core/os/execution/deprecated_execute_entry_point.cairo
+++ b/src/starkware/starknet/core/os/execution/deprecated_execute_entry_point.cairo
@@ -128,7 +128,7 @@ func deprecated_execute_entry_point{
         // Assert that there is no call data in the case of NOP entry point.
         assert execution_context.calldata_size = 0;
         %{ execution_helper.skip_call() %}
-        return (retdata_size=0, retdata=cast(0, felt*));
+        return (retdata_size=0, retdata=cast(nondet %{ segments.add() %}, felt*));
     }

     local range_check_ptr = range_check_ptr;
diff --git a/src/starkware/starknet/core/os/execution/deprecated_execute_syscalls.cairo b/src/starkware/starknet/core/os/execution/deprecated_execute_syscalls.cairo
index a3d2f922ec..da1c5219ea 100644
--- a/src/starkware/starknet/core/os/execution/deprecated_execute_syscalls.cairo
+++ b/src/starkware/starknet/core/os/execution/deprecated_execute_syscalls.cairo
@@ -248,7 +248,7 @@ func execute_deploy_syscall{
     assert syscall_ptr.response = DeployResponse(
         contract_address=contract_address,
         constructor_retdata_size=0,
-        constructor_retdata=cast(0, felt*),
+        constructor_retdata=cast(nondet %{ segments.add() %}, felt*),
     );

     tempvar constructor_execution_context = new ExecutionContext(
diff --git a/src/starkware/starknet/core/os/execution/execute_entry_point.cairo b/src/starkware/starknet/core/os/execution/execute_entry_point.cairo
index ef2ca40441..7b32e759af 100644
--- a/src/starkware/starknet/core/os/execution/execute_entry_point.cairo
+++ b/src/starkware/starknet/core/os/execution/execute_entry_point.cairo
@@ -171,7 +171,7 @@ func execute_entry_point{
         // Assert that there is no call data in the case of NOP entry point.
[...]

PR #428, opened 2024-11-07
No deploy nullptrs (except deprecated)

notlesh:
This is the same as #427 except that the OS still uses `cast(0, felt*)` during deprecated deploys.


notlesh:
Closing as this issue was addressed [in cairo-vm](https://github.com/lambdaclass/cairo-vm/pull/1862)

Issue #429, opened 2024-11-08
bug: `Block 174731` - `ids.a` Expected as Integer


whichqua:
### **Description**  
Fixing block 174731 involves addressing an exception that expects `ids.a` to be an integer value. The issue is triggered when calling the `get_tx_info()` function inside `__validate_deploy__`. The following code block replicates the issue:

```cairo
@raw_input
@external
func __validate_deploy__{
    syscall_ptr: felt*,
    pedersen_ptr: HashBuiltin*,
    ecdsa_ptr: SignatureBuiltin*,
    range_check_ptr
} (selector: felt, calldata_size: felt, calldata: felt*) {
    alloc_locals;
    // get the tx info
    let (tx_info) = get_tx_info();
    with_attr error_message("assertion failed") {
        assert_nn(tx_info.signature_len - 2);
    }

    return ();
}
```

### **Observed Behavior**  
The bug manifests in two primary ways:
1. **Direct Call to `get_tx_info()`**  
   - Executing `let (tx_info) = get_tx_info();` results in an error:
   ```
   DiffAssertValues((RelocatableValue(Relocatable { segment_index: 60, offset: 0 }), RelocatableValue(Relocatable { segment_index: 1, offset: 13281 }))
   ```
2. **Accessing `tx_info` Properties**  
   - Attempting to access `tx_info` (e.g., `assert_nn(tx_info.signature_len - 2);`) raises another error:
   ```
   Hint((0, IdentifierNotInteger("a")))
   ```
   - This appears related to execution order, where a hint tries to access a memory segment that isn’t valid anymore.

### **Affected Areas**  
This bug impacts both direct `deploy_account` calls and proxy contract calls.
This issue requires a deep dive into the `__validate_deploy__` function and `get_tx_info()` to align the expected data types and memory segments to avoid the described errors.

whichqua:
I can also confirm that calling `get_tx_signature` inside `__validate_deploy__` causes problems:
```cairo
@raw_input
@external
func __validate_deploy__{
    syscall_ptr: felt*,
    pedersen_ptr: HashBuiltin*,
    ecdsa_ptr: SignatureBuiltin*,
    range_check_ptr
} (selector: felt, calldata_size: felt, calldata: felt*) {
    alloc_locals;
    let (tx_i, tx_s ) = get_tx_signature();
    return ();
}
```
Results into this error:
```
[...]

PR #431, opened 2024-11-08
fix:`Block 174731` - `ids.a` Expected as Integer

whichqua:
Issue Number: #429

## Type

- [ ] feature
- [x] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Description
The issue is triggered when calling the `get_tx_info()` function inside `__validate_deploy__`.
The steps to fix the issue:
- [x] Replicate the issue with a simple integration test
- [x] Find the root of the issue
- [x] Write a fix
- [x] Sepolia test

## Breaking changes?

- [ ] yes
- [x] no


tests/integration/contracts/blockifier_contracts/feature_contracts/cairo0/test_proxy.cairo
```diff
@@ -0,0 +1,112 @@
+%lang starknet
+
+from starkware.cairo.common.cairo_builtins import HashBuiltin
+from starkware.starknet.common.syscalls import library_call, library_call_l1_handler
+from starkware.cairo.common.math import assert_not_zero
+
```
notlesh, line 6: If there's any context you can add about this file (where it came from, why we have it, etc.) it would be good to add here.

tests/integration/contracts/blockifier_contracts/feature_contracts/cairo0/test_proxy_account.cairo
```diff
@@ -0,0 +1,98 @@
...
+} (selector: felt, calldata_size: felt, calldata: felt*) {
+    alloc_locals;
+    // get the tx info
+    let (tx_info) = get_tx_info();
+    with_attr error_message("assertion failed") {
+        assert_nn(tx_info.signature_len - 2);
+    }
+
+    return ();
+}
```
notlesh, line 90: Is this still needed? If we don't need all of the unusual things here (e.g. @raw_input) in order to test this (as a regression test) then I would suggest removing it or at least adding some documentation about it. Otherwise I can see it adding a lot of confusion in the future.

tests/integration/deploy_txn_tests.rs
```diff
@@ -358,3 +358,144 @@ async fn deploy_via_invoke_no_calldata_cairo1_account(
...
+        initial_state.cairo1_compiled_classes,
+        HashMap::default(),
+    )
+    .await
+    .expect("OS run failed");
+}
+
+#[rstest]
+#[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+async fn deploy_cairo0_via_proxy(block_context: BlockContext, max_fee: Fee) {
```
notlesh, line 429: Does the proxy impact this bug at all? I might apply my same concerns above if not, this is likely to add confusion in the future.

crates/starknet-os/src/hints/mod.rs
```diff
@@ -626,8 +626,7 @@ pub async fn start_tx_async<PCS>(
 where
     PCS: PerContractStorage + 'static,
 {
-    let deprecated_tx_info_ptr =
-        get_relocatable_from_var_name(vars::ids::DEPRECATED_TX_INFO, vm, ids_data, ap_tracking)?;
+    let deprecated_tx_info_ptr = get_ptr_from_var_name(vars::ids::DEPRECATED_TX_INFO, vm, ids_data, ap_tracking)?;
```
notlesh, line 629: image

notlesh (commented):
LGTM, but I think we should clean up some of the test code that isn't relevant

crates/starknet-os/src/hints/mod.rs
HermanObst: HAHAHAHAHAHAHAHHA ❤️

tests/integration/contracts/blockifier_contracts/feature_contracts/cairo0/test_proxy_account.cairo
whichqua: Resolved by https://github.com/keep-starknet-strange/snos/pull/431/commits/c4726a7c44e2a64d7156aeaea7075d52c14ee601

tests/integration/deploy_txn_tests.rs
whichqua: Resolved by https://github.com/keep-starknet-strange/snos/pull/431/commits/c4726a7c44e2a64d7156aeaea7075d52c14ee601

crates/starknet-os/src/hints/tests.rs
```diff
@@ -158,7 +158,9 @@ pub mod tests {
         vm.add_memory_segment();
         vm.add_memory_segment();
 
-        let ids_data = ids_data![vars::ids::DEPRECATED_TX_INFO];
+        let mut ids_data = HashMap::new();
+        ids_data.insert(vars::ids::DEPRECATED_TX_INFO.to_owned(), HintReference::new(0, 0, false, false));
+
```
HermanObst, line 163: @whichqua why this change is needed?

whichqua: Because the vm treats `pointers` and `relocalables` differently. This is how to define pointers.

tests/integration/deploy_txn_tests.rs
```diff
@@ -358,3 +358,63 @@ async fn deploy_via_invoke_no_calldata_cairo1_account(
...
     .expect("OS run failed");
 }
+
+#[rstest]
+#[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+async fn deploy_cairo0_check_get_info_call(block_context: BlockContext, max_fee: Fee) {
+    let account_with_tx_info_check = load_cairo0_feature_contract("account_with_tx_info_check");
+    let class_hash = class_hash!("0x6c8903651a5f89ffc304621a7d8106a0324cc28aca04934fcbbb4398d5c8bc8");
+
+    let args: Vec<Felt252> = [].into_iter().map(Felt252::from_hex_unchecked).collect();
```
HermanObst, line 368: Why not just?

`let args: Vec<Felt252> = Vec::new();
`

HermanObst (changes requested):

HermanObst (approved):
LGTM!

ftheirs (approved):
This looks awesome 😎

PR #432, opened 2024-11-12
Improve README

HermanObst:
As we are approaching to our first release, we need a better README that makes SNOS easier for everyone to use.


.gitignore
```diff
@@ -12,6 +12,8 @@ build/*
 # Cairo dependencies
 cairo-lang
 venv
+snos-env
+.python-version
```
shamsasari, line 16: Why is this being ignored?

README.md
```diff
@@ -1,38 +1,155 @@
...
+- [🤝 Related Projects](#-related-projects)
+- [📚 Documentation](#-documentation)
+- [📜 License](#-license)
+
+
+## 📖 About
+
+[Starknet OS](https://github.com/starkware-libs/cairo-lang/blob/master/src/starkware/starknet/core/os/os.cairo) is a [Cairo](https://www.cairo-lang.org/) program designed to prove the integrity of state transitions between blocks on Starknet.
+
+By re-executing transactions from a block and verifying consistency, it produces a [PIE](https://github.com/starkware-libs/cairo-lang/blob/a86e92bfde9c171c0856d7b46580c66e004922f3/src/starkware/cairo/lang/vm/cairo_pie.py#L219-L225) (Program Independent Execution) result. This PIE can be used to generate a Stark proof of integrity, which, if accepted by Starknet L1 verifiers, confirms block validity and updates the Starknet state root in the [StarknetCore contract](https://etherscan.io/address/0xc662c410c0ecf747543f5ba90660f6abebd9c8c4#code).
```
shamsasari, line 39: Should it be `STARK` not `Stark`?

shamsasari, line 39: Can you add 120 (or whatever the amount) char hardwraps

README.md
```diff
@@ -1,38 +1,155 @@
...
+## 🛠️ Getting Started
 
-## Test Setup
+### Prerequisites
 
-**Cairo Env**
-(see: https://docs.cairo-lang.org/0.12.0/quickstart.html)
+Ensure you have the following dependencies installed:
+- [Rust 1.76.0 or newer](https://www.rust-lang.org/tools/install)
+- Cargo (included with Rust)
```
shamsasari, line 47: Redundant

shamsasari (changes requested):

README.md
```diff
@@ -1,38 +1,155 @@
...
+## Table of Contents
+- [Table of Contents](#table-of-contents)
+- [📖 About](#-about)
+- [🛠️ Getting Started](#️-getting-started)
+  - [Prerequisites](#prerequisites)
+  - [Installation](#installation)
+- [🧪 Running Tests](#-running-tests)
+  - [Run Tests](#run-tests)
+  - [Reset Tests](#reset-tests)
+- [🚀 Usage](#-usage)
```
whichqua, line 27: Maybe Usage should come before `Running Tests` as most users will possibly be looking to use the lib/bin rather than fork the repo and run the tests.

README.md
```diff
@@ -1,38 +1,155 @@
...
+
+The `prove_block` binary handles this entire process by collecting, formatting, and feeding the necessary data into the OS, ensuring the correct `OSInput` is passed for execution.
+
+To accomplish this, it queries the required information from a full node. Currently, Pathfinder is the only full node implementing all the necessary RPC methods, so a synced [Pathfinder](https://github.com/eqlabs/pathfinder) instance running as an [**archive node**](https://github.com/eqlabs/pathfinder?tab=readme-ov-file#state-trie-pruning) (to provide access to storage proofs) is required to execute this binary successfully.
+
+For example, you can run Pathfinder executing:
+
+```bash
+PATHFINDER_ETHEREUM_API_URL="YOUR_KEY" ./target/release/pathfinder --data-directory /home/herman/pathfinder-data --http-rpc 0.0.0.0:9545 --storage.state-tries archive
```
whichqua, line 118: Might it be better to give the docker command instead of the binary? This would mean I can copy the command one time and just run it to setup pathfinder/

whichqua (commented):
Left some few comments, let me know your thoughts.

.gitignore
HermanObst: Because that is stuff generated by the setup

setup-scripts/reset-tests.sh
```diff
@@ -2,5 +2,5 @@
 
 bash scripts/teardown-tests.sh
 
-bash scripts/setup-tests.sh
+bash setup-scripts/setup-tests.sh
```
notlesh, line 5: I don't think we need both `scripts/` and `setup-scripts/`, maybe put the setup ones inside a subdir of `scripts`? e.g.:

```
scripts/
scripts/<remaining shell scripts>
scripts/setup_scripts/
scripts/setup_scripts/<moved scripts>
```

README.md
```diff
@@ -1,38 +1,155 @@
...
 ./scripts/reset-tests.sh
 ```
+
+##  🚀 Usage 
+### Adding SNOS as a dependency
+
+You can add the following to your rust project's `Cargo.toml`:
+
+```toml
+starknet-os = { git = "https://github.com/keep-starknet-strange/snos", rev = "662d1706f5855044e52ebf688a18dd80016c8700" }
```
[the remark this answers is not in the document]
HermanObst: we can add that when we add SNOS into crates.io
For now this is more as a placeholder

README.md
HermanObst: Defer to later as this triggers much deeper conversations
(basically how to package SNOS and the os.cairo)

setup-scripts/reset-tests.sh
HermanObst: Defer to later as I prefer users to not have to interact with scripts under `/scripts`

.gitignore
shamsasari: I was referring to `.python-version` - it fixes the python version to use. Is there not a specific version of python that must be used? Your setup-cairo script is mandating verion 3.9.15.

README.md
shamsasari: It's installed as standard - the link you provide will install rustup and also cargo.

README.md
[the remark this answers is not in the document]
shamsasari: > we can add that when we add SNOS into crates.io

When will that be?

It takes 2 seconds to create a tag at that commit. Otherwise it doesn't look very professional to ask users to use a "random" commit.

README.md
```diff
@@ -1,38 +1,155 @@
...
+```
+
+### Using the `prove_block` Binary
```
shamsasari, line 103: also, binary shouldn't be capitalised here. Sorry if this is nit-picking, but good grammer and English goes a long way in any documentation.

README.md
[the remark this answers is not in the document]
HermanObst: I would say that it makes sense to add a tag and point to a specific commit when we finish debugging and have 100% success in Sepolia.
This will happen in around 3 weeks and its when we will make our first release

shamsasari (approved):

Issue #434, opened 2024-11-19
bug: Block 161599 - Dict Error: No value found for key


ftheirs:
**Description:**
After running SNOS for block `161599`, the following error occurred:

```
inner_exc error: Got an exception while executing a hint: Dict Error: No value found for key: 1437646471060246884922185914160069087116232223052640763259807059700780118310
```

This error suggests that [this function call](https://github.com/starkware-libs/cairo-lang/blob/8e11b8cc65ae1d0959328b1b4a40b92df8b58595/src/starkware/starknet/core/os/execution/execute_entry_point.cairo#L149-L151) is failing due to a missing key.

Upon further inspection, the missing keys can be found in the [storage_proofs](https://github.com/keep-starknet-strange/snos/blob/9706b8aad0932e5564d90c85dbb9898d90b5b61b/crates/bin/prove_block/src/lib.rs#L188-L190), but they are absent from the [traces](https://github.com/keep-starknet-strange/snos/blob/9706b8aad0932e5564d90c85dbb9898d90b5b61b/crates/bin/prove_block/src/state_utils.rs#L47-L48).

If you examine the failing transaction, it is the first transaction in the block, and it has been reverted. Closer analysis of the traces reveals that execute_invocation is only processed when the transaction result is Success:

``` Rust
TransactionTrace::Invoke(invoke_trace) => {
    if let Some(inv) = &invoke_trace.validate_invocation {
        process_function_invocations(inv, &mut contracts_subcalled, &mut classes_subcalled);
    }
    if let ExecuteInvocation::Success(inv) = &invoke_trace.execute_invocation {
        process_function_invocations(inv, &mut contracts_subcalled, &mut classes_subcalled);
    }
    if let Some(inv) = &invoke_trace.fee_transfer_invocation {
        process_function_invocations(inv, &mut contracts_subcalled, &mut classes_subcalled);
    }
}
```

For reverted transactions, a revert_error is extracted instead of execution_invocation. In this case, the error was:

```
[...]

PR #436, opened 2024-11-20
Fix reexecute with blockifier

ftheirs:
Problem:
Certain blocks encountered different errors when processed with SNOS. Upon closer inspection, discrepancies were identified between the transaction traces provided by Pathfinder and the internal re-execution using Blockifier. Since both components essentially utilize the same parts of Blockifier, and the traces accurately reflect the network's state, we determined that the traces could serve as a source of truth.

Solution:
Identify the differences in how Pathfinder executes transactions using Blockifier and apply the necessary adjustments. This PR calls `pre_process_block` before executing the transactions that was missing.

Closes https://github.com/keep-starknet-strange/snos/issues/434
Issue Number: N/A

Fix blocks: `161599`, `174156`, `161884` and `174027`

## Type

- [ ] feature
- [x] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [ ] yes
- [x] no

### DIFF
--- a/crates/bin/prove_block/src/reexecute.rs
+++ b/crates/bin/prove_block/src/reexecute.rs
@@ -41,14 +41,14 @@ fn get_tx_hash(tx: &Transaction) -> TransactionHash {
 pub fn reexecute_transactions_with_blockifier<S: StateReader>(
     state: &mut CachedState<S>,
     block_context: &BlockContext,
-    previous_block_hash: Felt252,
+    buffer_block_hash: Felt252,
     txs: Vec<Transaction>,
 ) -> Result<Vec<TransactionExecutionInfo>, Box<dyn Error>> {
     let current_block_number = block_context.block_info().block_number;
-    let previous_block_number_and_hash = if current_block_number.0 >= STORED_BLOCK_HASH_BUFFER {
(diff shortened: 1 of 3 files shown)


crates/bin/prove_block/src/reexecute.rs
```diff
@@ -40,10 +41,25 @@ fn get_tx_hash(tx: &Transaction) -> TransactionHash {
 pub fn reexecute_transactions_with_blockifier<S: StateReader>(
     state: &mut CachedState<S>,
     block_context: &BlockContext,
+    previous_block_hash: Felt252,
```
HermanObst, line 44: Dont call this `previous_block_hash`. 
Is misleading as we are calling previous to current - 1 block

`old_block_hash` its ok. But  `buffer_block_hash` would be ideal

HermanObst (changes requested):
Minor change. The rest looks great

HermanObst (approved):
NICE, LGTM!!!!!!!

whichqua (approved):
LGTM!
Good work @ftheirs

GMKrieger (approved):
Great work, Fer!

PR #440, opened 2024-11-26
Remove unused code from starkware_utils

GMKrieger:
Issue Number: N/A

## Type

- [ ] feature
- [ ] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Description

Removed several unused methods and unreachable code from `starkware_utils`.

## Breaking changes?

- [ ] yes
- [x] no


crates/starknet-os/src/starkware_utils/commitment_tree/binary_fact_tree_node.rs
```diff
@@ -174,20 +174,7 @@ where
         ffc: &'a mut FactFetchingContext<S, H>,
         indices: &'a [TreeIndex],
         facts: &'a mut Option<BinaryFactDict>,
-    ) -> impl std::future::Future<Output = Result<HashMap<TreeIndex, LF>, TreeError>> + Send {
```
whichqua, line 177: Any reason why you have removed this default implementation?

whichqua (commented):
Left a single comment but most seems good.

crates/starknet-os/src/starkware_utils/commitment_tree/binary_fact_tree_node.rs
GMKrieger: No particular reason except it's never used. This trait is only implemented in `virtual_patricia_node.rs`, and it has it's own implementation of _get_leaves.

ftheirs (approved):
This looks good 💪

notlesh (commented):
A lot of what is removed here was from the original port of the pythonic impl to rust. That was done in as much of a literal fashion as was possible, and that's probably why there is a lot of dead code.

I can see arguments for and against keeping this code, so perhaps we should discuss internally before we decide.

whichqua (approved):
LGTM!

HermanObst (approved):
LGTM

PR #442, opened 2024-11-28
Key not found in preimage

### LINKED ISSUE #443: bug: Key not found in preimage
**Issue Description**
While running SNOS for blocks `237025, 237030, 237037, 237042, 237044, 237053, 237083, 237086, 235385, 235620`, the following error consistently occurred:

``` vbnet
[ERROR prove_block]  
    inner_exc error: Got an exception while executing a hint: Hint Error: Key not found in preimage: 1496700206899231316563006125976608562510487632958670553981981720152324726300  
```

The error varies depending on the block being processed, but it always fails at the same spot in the execution flow.

-----------------------------------------

**Debugging Steps and Observations**
1- Tracing the Error:
Upon analyzing the trace, it became evident that the error originates in [this hint](https://github.com/keep-starknet-strange/snos/blob/cb2a6d26faeb658492756fe100bbdf5b1600c768/crates/starknet-os/src/hints/patricia.rs#L311-L362), which fails when a requested node is missing from the preimage. The error indicates that a node required for the proof is not present.

2- Inspecting the Preimage Initialization:
The next logical step was to trace how the preimage is set up. Following the SNOS workflow, it was found that [this hint](https://github.com/keep-starknet-strange/snos/blob/cb2a6d26faeb658492756fe100bbdf5b1600c768/crates/starknet-os/src/hints/state.rs#L48-L80) initializes the preimage using `os_input.contract_state_commitment_info.commitment_facts`.

3- Tracking the Source of Commitment Facts:
Digging further, the `commitment_facts` are derived from the [storage_proofs](https://github.com/keep-starknet-strange/snos/blob/cb2a6d26faeb658492756fe100bbdf5b1600c768/crates/bin/prove_block/src/lib.rs#L315-L319), specifically within the get_storage_proof function. This function is defined [here](https://github.com/keep-starknet-strange/snos/blob/cb2a6d26faeb658492756fe100bbdf5b1600c768/crates/bin/prove_block/src/rpc_utils.rs#L114-L139).

4- Identifying the Missing Nodes:
Through debugging, it was discovered that the missing nodes belong to the storage proof for contract address `0x1`. This is a special contract from Starknet's architecture that maps block numbers to their corresponding block hashes. However, the connection between the transactions and the error wasn’t immediately apparent.

5- Investigating the Triggering Contract:
Further investigation revealed that contract `0x00656ca4889a405ec5222e4b0997e5a043902a98cb1f85a039f76f50c000479d` is responsible for triggering the issue. The problem occurs in specific functions, such as `verify_game` and `abandon_game`. Upon inspecting the Cairo code for this contract, these functions were found to internally call [a utility function](https://github.com/Await-0x/dark-shuffle/blob/8211e187a4e872bc567385921c7dda6276110aad/contracts/src/systems/game/contracts.cairo#L82-L101), which in turn calls `get_block_hash_syscall` (more documentation can be found [here](https://docs.starknet.io/architecture-and-concepts/smart-contracts/system-calls-cairo1/#get_block_hash))

-----------------------------------------

**Conclusion and Solution**
The error occurs because the storage proof for contract 0x1 does not include all the necessary keys for the calls and subcalls executed during the transaction. To resolve this issue:

1- Add Missing Keys:
Ensure that all necessary keys for the storage proof of contract `0x1` are included before execution. This involves analyzing the calls and subcalls from all the transactions and identifying the keys they depend on.

2- Optimize the Key Retrieval Process:
Since the exact keys needed cannot always be determined upfront, the retrieval process has been optimized. Basically, we leverage from the structure from this special contract and filter out any keys with values higher than the current block number. This minimizes unnecessary data while ensuring all relevant nodes are included in the preimage.

ftheirs:
Problem: Execution from SNOS for the following blocks results in a `key not found in preimage` error:
`237025, 237030, 237037, 237042, 237044, 237053, 237083, 237086, 235385, 235620`
Better description from the issue can be found in #443 

Solution:
Add the missing keys required to retrieve the correct storage proof for the contract address `0x1`.

Closes #443 

## Type

- [ ] feature
- [x] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [ ] yes
- [x] no


crates/bin/prove_block/src/rpc_utils.rs
```diff
@@ -111,16 +111,73 @@ fn verify_storage_proof(contract_data: &ContractData, keys: &[Felt]) -> Vec<Felt
...
+/// The first approach would introduce significant overhead for most blocks, and the second solution is currently not feasible at the moment.
+fn insert_extra_storage_reads_keys(
+    storage_reads: &HashMap<ContractAddress, Vec<Felt252>>,
+    old_block_number: Felt252,
+    keys: &mut HashMap<ContractAddress, HashSet<StorageKey>>,
+) {
+    // A list of the contracts that accessed to the storage from 0x1 using `get_block_hash_syscall`
+    let special_addresses: Vec<ContractAddress> = vec![
+        contract_address!("0x01246c3031c5d0d1cf60a9370aac03a4717538f659e4a2bfb0f692e970e0c4b5"),
+        contract_address!("0x00656ca4889a405ec5222e4b0997e5a043902a98cb1f85a039f76f50c000479d"),
```
notlesh, line 127: Do you know how these hashes are derived? I think this might help us know exactly where this problem could be fixed.

crates/bin/prove_block/src/rpc_utils.rs
```diff
@@ -111,16 +111,73 @@ fn verify_storage_proof(contract_data: &ContractData, keys: &[Felt]) -> Vec<Felt
...
+/// This approach serves as a workaround. The ideal solutions would be to either retrieve the full tree or obtain information about the necessary nodes.
+/// The first approach would introduce significant overhead for most blocks, and the second solution is currently not feasible at the moment.
+fn insert_extra_storage_reads_keys(
+    storage_reads: &HashMap<ContractAddress, Vec<Felt252>>,
+    old_block_number: Felt252,
+    keys: &mut HashMap<ContractAddress, HashSet<StorageKey>>,
+) {
+    // A list of the contracts that accessed to the storage from 0x1 using `get_block_hash_syscall`
+    let special_addresses: Vec<ContractAddress> = vec![
+        contract_address!("0x01246c3031c5d0d1cf60a9370aac03a4717538f659e4a2bfb0f692e970e0c4b5"),
```
whichqua, line 126: @ftheirs Will these contract addresses apply on mainnet?

ftheirs: You're right! These addresses correspond to contracts deployed on Sepolia. For sure we'll have similar issues in mainnet

crates/bin/prove_block/src/rpc_utils.rs
```diff
@@ -111,16 +111,75 @@ fn verify_storage_proof(contract_data: &ContractData, keys: &[Felt]) -> Vec<Felt
...
+        contract_address!("0x00656ca4889a405ec5222e4b0997e5a043902a98cb1f85a039f76f50c000479d"),
+        // Ekubo:core contract address. Source code is not available but `key_not_in_preimage` error is triggered every time it's called
+        contract_address!("0x5dd3d2f4429af886cd1a3b08289dbcea99a294197e9eb43b0e0325b4b"),
+    ];
+
+    let extra_storage_reads = if special_addresses.iter().any(|address| storage_reads.contains_key(address)) {
+        100 * STORED_BLOCK_HASH_BUFFER
+    } else {
+        STORED_BLOCK_HASH_BUFFER
+    };
```
HermanObst, line 136: x100 multiplier fix everything (also Mainnet?)

crates/bin/prove_block/src/rpc_utils.rs
```diff
@@ -111,16 +111,75 @@ fn verify_storage_proof(contract_data: &ContractData, keys: &[Felt]) -> Vec<Felt
...
+        // Include not only old_block_hash but also retrieve the previous 10 values
+        let block_hash_contract_address = contract_address!("0x1");
+        keys.entry(block_hash_contract_address)
+            .or_default()
+            .insert((old_block_number).try_into().expect("Felt to StorageKey conversion failed"));
+
+        // Include extra keys for contracts that trigger get_block_hash_syscall
+        insert_extra_storage_reads_keys(&storage_reads, old_block_number, &mut keys);
+
+        // Within the Starknet architecture, the address 0x1 is a special address that maps block numbers to their corresponding block hashes. As some contracts might access storage reads using `get_block_hash_syscall`, it is necessary to add some extra keys here.
```
HermanObst, line 166: As some contracts might access storage reads using `get_block_hash_syscall` and Blockifier currently do not stores these accessed values.

crates/bin/prove_block/src/rpc_utils.rs
```diff
@@ -111,16 +111,75 @@ fn verify_storage_proof(contract_data: &ContractData, keys: &[Felt]) -> Vec<Felt
...
+        // By leveraging the structure of this special contract, we filter out storage_read_values that are greater than old_block_number and add these extra values to the necessary keys from the 0x1 contract address.
+        // It is worth noting that this approach incurs some overhead due to the retrieval of additional data.
+        // let additional_storage_reads: Vec<Felt> = storage_reads.values().flat_map(|vec| vec.iter()).cloned().collect();
+        let additional_storage_reads: Vec<Felt> = storage_reads
+            .values()
+            .flat_map(|vec| vec.iter().cloned().filter(|x| x <= &Felt252::from(block_number)))
+            .collect::<HashSet<Felt>>()
+            .into_iter()
+            .collect();
```
HermanObst, line 175: What is the rationale for this?
It seems that you are preventing to call storage values (of contract 0x1) that are also called in another contract tree. 

Why just not get all the storage values that we will ask (we obtain this in `insert_extra_storage_reads_keys`, given if we need it or not) and just extend 
`keys.entry(block_hash_contract_address).or_default().extend(`

It seems that we are adding, subtracting and doing some logic that could be much simpler. 

Does it makes sense?

HermanObst (changes requested):
Hey there! Good work on this :) 
I just added some questions as it seems the logic can be much simpler!

crates/bin/prove_block/src/rpc_utils.rs
```diff
@@ -111,16 +111,75 @@ fn verify_storage_proof(contract_data: &ContractData, keys: &[Felt]) -> Vec<Felt
     additional_keys
 }
 
+/// Inserts additional keys for retrieving storage proof from the block hash contract (address 0x1).
+/// Certain contracts necessitate extra nodes from the contract 0x1. However, since Blockifier does not provide this information,
+/// it is necessary to add some extra keys to ensure the inclusion of the required nodes.
+/// This approach serves as a workaround. The ideal solutions would be to either retrieve the full tree or obtain information about the necessary nodes.
```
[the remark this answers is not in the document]
HermanObst: ping on this

HermanObst (approved):
LGTM

PR #445, opened 2024-12-04
Pre-release versioning

notlesh:
## Description

This PR prepares SNOS for its first release by bumping each crate's version to `1.0.0-dev`. It's also a suggestion on how to implement versioning based on [Semver](https://semver.org/).

Each crate is independently versioned. This is so changes to one crate don't create noise for the users of other crates.

I also suggest adopting `-dev` tag to indicate a pre-release version. This is inspired by other projects like [Frontier](https://github.com/polkadot-evm/frontier?tab=readme-ov-file#versioning). It's explained well there, but to summarize: appending `-dev` to the version string flags the version as in the process of being bumped in the next release, making it easier to understand when and how a PR should bump the version.

Issue Number: N/A

## Type

- [ ] feature
- [ ] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [ ] yes
- [x] no


HermanObst (changes requested):
@notlesh some thoughts:
- I would change the version to v0.1.0 (not v1.0.0)
- The cairo-lang submodule shouldn't be modified

PR #447, opened 2024-12-08
Refactor/env utils

heemankv:
Issue Number: N/A

## Type

- [ ] feature
- [ ] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Description
- This PR allows for env based input of certain variables.
- Currently being used for `strk` and `eth` token addresses found in `rpc-replay` crate. 
- While testing sequencers such as [Madara](https://github.com/madara-alliance/madara) with [Madara Orchestrator](https://github.com/madara-alliance/madara-orchestrator), having this capability to customise the address allows for testing with dummy tokens, which the tester can control more easily.

## Breaking changes?

- [ ] yes
- [x] no


crates/utils/src/env_utils.rs
```diff
@@ -0,0 +1,27 @@
...
+    match get_env_var(key) {
+        // if value is empty string, return None
+        Ok(s) if s.is_empty() => Ok(None),
+        Ok(s) => Ok(Some(s)),
+        Err(VarError::NotPresent) => Ok(None),
+        Err(e) => Err(e),
+    }
+}
+
+pub fn get_env_car_optional_or_panic(key: &str) -> Option<String> {
```
apoorvsadana, line 25:
```suggestion
pub fn get_env_var_optional_or_panic(key: &str) -> Option<String> {
```

crates/utils/src/env_utils.rs
apoorvsadana, line 1: should we keep all if they aren't being used right now?

crates/utils/src/env_utils.rs
heemankv: Function removed.

crates/utils/src/env_utils.rs
heemankv: Valid, we'll add them when we need them.
Removed for now.

crates/utils/src/lib.rs
```diff
@@ -0,0 +1 @@
+pub mod env_utils;
```
whichqua, line 1: Do we need a full crate for this? cc @notlesh

whichqua: @heemankv I would suggest so. Or just inline the code and skip the functions part.

heemankv: Alright added, can you re-review @whichqua

crates/rpc-replay/src/block_context.rs
```diff
@@ -11,6 +12,12 @@ use starknet_api::{contract_address, felt, patricia_key};
 
 use crate::utils::{felt_to_u128, FeltConversionError};
 
+/// Fetches environment variable based on provided key.
+/// If key is not found in env, it returns default value provided
+pub fn get_env_var_or_default(key: &str, default: &str) -> String {
```
whichqua, line 17:
```suggestion
#[inline]
fn get_env_var_or_default(key: &str, default: &str) -> String {
```

crates/rpc-replay/Cargo.toml
```diff
@@ -16,6 +16,7 @@ starknet_api = { workspace = true }
 starknet-os-types = { workspace = true }
 thiserror = { workspace = true }
 tokio = { workspace = true }
+serial_test = "3.2.0"
```
whichqua, line 19: This is a dev dependency. +1 if we can avoid adding it

crates/rpc-replay/src/block_context.rs
```diff
@@ -49,16 +56,21 @@ pub fn build_block_context(
...
         chain_id,
         // cf. https://docs.starknet.io/tools/important-addresses/
         fee_token_addresses: FeeTokenAddresses {
-            strk_fee_token_address: contract_address!(
-                "0x04718f5a0fc34cc1af16a1cdee98ffb20c31f5cd61d6ab07201858f4287c938d"
-            ),
-            eth_fee_token_address: contract_address!(
-                "0x049d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7"
-            ),
+            strk_fee_token_address: contract_address!(strk_fee_token_address.as_str()),
```
whichqua, line 72: Do we need `.as_str()` here? Adding a reference means `&String` will deref to `&str`
```suggestion
            strk_fee_token_address: contract_address!(&strk_fee_token_address),
```

crates/rpc-replay/src/block_context.rs
```diff
@@ -70,12 +82,34 @@ pub fn build_block_context(
 
 #[cfg(test)]
 mod tests {
-
     use starknet::core::types::{Felt, ResourcePrice};
     use starknet_api::core::ChainId;
 
     use super::*;
 
+    #[test]
```
whichqua, line 90: Wont these tests interfere with each other if run in parallel?

crates/rpc-replay/src/block_context.rs
```diff
@@ -49,16 +56,21 @@ pub fn build_block_context(
         use_kzg_da,
     };
 
+    let strk_fee_token_address = get_env_var_or_default(
+        "SNOS_STRK_FEE_TOKEN_ADDRESS",
+        "0x04718f5a0fc34cc1af16a1cdee98ffb20c31f5cd61d6ab07201858f4287c938d",
+    );
+    let eth_fee_token_address = get_env_var_or_default(
+        "SNOS_ETH_FEE_TOKEN_ADDRESS",
+        "0x049d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7",
```
whichqua, line 65: Consider creating constants for these, something like:
```
const DEFAULT_ETH_FEE_TOKEN_ADDRESS = ....
```

Cargo.toml
whichqua, line 1: Are these changes needed?

whichqua (changes requested):
Left some needed changes. Let me know if you have any thoughts or questions.

crates/rpc-replay/src/block_context.rs
heemankv:
```rust
contract_address!(&eth_fee_token_address)
```
```
the trait bound `FeltConverter: TryIntoFelt<&String>` is not satisfied
the following other types implement trait `TryIntoFelt<V>`:
  `FeltConverter` implements `TryIntoFelt<&str>`
  `FeltConverter` implements `TryIntoFelt<u128>`
  `FeltConverter` implements `TryIntoFelt<u16>`
  `FeltConverter` implements `TryIntoFelt<u32>`
  `FeltConverter` implements `TryIntoFelt<u64>`
  `FeltConverter` implements `TryIntoFelt<u8>`
```

IMO yes we need this, can you think of any other approach ?

crates/rpc-replay/src/block_context.rs
heemankv: Yes, and they do interfere, hence I added the `serial_test`,
I suggest any of the two options here : 
- We rework this test case to only test the env utils and not test integration of env utils in the `build_block_context` function.
- We keep `serial_test` to ensure that serialisation.
WDYT ?

crates/rpc-replay/src/block_context.rs
heemankv: Valid, added.

Cargo.toml
heemankv: These are lint changes, will revert to keep as is.

crates/rpc-replay/src/block_context.rs
heemankv: Testing with ` cargo test -- --nocapture --test-threads=3 ` works 
image

No they don't since they are using different `test_key`

crates/rpc-replay/Cargo.toml
heemankv: I have now updated it to be a dev-dependency, I hope that solves the issue,
This is required to ensure 
- `test_build_block_context_with_default_env_works`
- `test_build_block_context_with_default_env_fails`
works together properly.
WDYT ?

crates/rpc-replay/src/block_context.rs
heemankv: Updated.

crates/rpc-replay/src/block_context.rs
whichqua: You can keep it as it is

whichqua (approved):
LGTM

notlesh (approved):

PR #448, opened 2024-12-09
fix minor issues in the script Update check-orphans.sh

mdqst:
## Type

- [ ] feature
- [x] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Description

1. **Removed unnecessary quotes around `0` in the condition check**:  
   In Bash, it's standard to avoid quotes around numeric values in conditions. This update replaces `if [ $? -ne '0' ]; then` with `if [ $? -ne 0 ]; then`, making the code more consistent with common Bash practices.

2. **Simplified `echo` command by removing `-e` flag**:  
   The `-e` flag in the `echo` command is used to interpret escape sequences, but there are no special characters to interpret in the string. Therefore, the flag was removed, and the code is now cleaner and more efficient.

## Breaking changes?

- [ ] yes
- [x] no

### DIFF
--- a/scripts/check-orphans.sh
+++ b/scripts/check-orphans.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 cargo run --quiet --example hints orphans | jq -e 'if length == 0 then . else empty end'
-if [ $? -ne '0' ]; then
-  echo -e "orphaned hints found:"
+if [ $? -ne 0 ]; then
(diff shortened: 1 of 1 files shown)


HermanObst:
Hey @mdqst thanks for the contribution. 
Im closing this PR as we are removing the check-orphan script

PR #454, opened 2024-12-27
log2 hint

ftheirs:
New hint from OS v0.13.3: log2 ceil. Implementation from the hint and a public function to compute the log2 since it will be needed in compression hints

KZG tests still require extra hints. 
Check orphan hints is failing because it detects two extra hints that will be covered in compress PR (`COMPRESS` and `SET_DECOMPRESSED_DST`)

Issue Number: N/A

## Type

- [x] feature
- [ ] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [ ] yes
- [x] no

### DIFF
--- a/crates/starknet-os/src/hints/math.rs
+++ b/crates/starknet-os/src/hints/math.rs
@@ -1,0 +1,119 @@
+use std::collections::HashMap;
+
+use cairo_vm::hint_processor::builtin_hint_processor::hint_utils::{
+    get_integer_from_var_name, insert_value_from_var_name,
+};
+use cairo_vm::hint_processor::hint_processor_definition::HintReference;
+use cairo_vm::serde::deserialize_program::ApTracking;
+use cairo_vm::types::exec_scope::ExecutionScopes;
+use cairo_vm::vm::errors::hint_errors::HintError;
+use cairo_vm::vm::vm_core::VirtualMachine;
+use cairo_vm::Felt252;
+use indoc::indoc;
+use num_bigint::BigUint;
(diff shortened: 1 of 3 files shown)


crates/starknet-os/src/hints/math.rs
```diff
@@ -0,0 +1,97 @@
...
+use num_bigint::BigUint;
+use num_traits::{One, Zero};
+
+use crate::hints::vars;
+
+pub const LOG2_CEIL: &str = indoc! {r#"
+	from starkware.python.math_utils import log2_ceil
+    ids.res = log2_ceil(ids.value)"#
+};
+pub fn log2_ceil_hint(
```
whichqua, line 22: We possibly just need to remove the hint from `unimplemented.rs`

whichqua (approved):
Should be good

crates/starknet-os/src/hints/unimplemented.rs
```diff
@@ -1,4 +1 @@
-use indoc::indoc;
-
-#[allow(unused)]
-pub const HINT_4: &str = indoc! {r#"exit_syscall(selector=ids.SHA256_PROCESS_BLOCK_SELECTOR)"#};
+// This file is intended to list all the unimplemented hints during an OS upgrade
```
notlesh, line 1:
```suggestion
// This file is intended to list all the unimplemented hints found by the `hint_tool` binary
// when comparing the hints found in the compiled OS compared to the hints known by SNOS
```

notlesh (approved):

PR #455, opened 2024-12-27
Dict hints

ftheirs:
Implement dictionary hints related to compression process.
Add constants.rs file to define constants that are needed across several hints

KZG tests still require extra hints. 
Check orphan hints is failing because it detects two extra hints that will be covered in compress PR (`COMPRESS` and `SET_DECOMPRESSED_DST`)

Issue Number: N/A

## Type

- [x] feature
- [ ] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [ ] yes
- [x] no


crates/starknet-os/src/hints/dict.rs
```diff
@@ -0,0 +1,137 @@
...
+    let dict_ptr = get_ptr_from_var_name(vars::ids::DICT_PTR, vm, ids_data, ap_tracking)?;
+
+    let dict_tracker = match exec_scopes.get_dict_manager()?.borrow().get_tracker(dict_ptr)?.data.clone() {
+        Dictionary::SimpleDictionary(hash_map) => hash_map,
+        Dictionary::DefaultDictionary { dict, .. } => dict,
+    };
+
+    let bucket_index = get_maybe_relocatable_from_var_name(vars::ids::BUCKET_INDEX, vm, ids_data, ap_tracking)?;
+
+    let prev_offset = dict_tracker.get(&bucket_index).unwrap().clone();
```
whichqua, line 55: Handling this unwrap is the only thing I can thing of.

whichqua (approved):
Should be good!

notlesh (changes requested):
The code looks good to me, but I would suggest one thing:

In `cairo-lang` this all comes from a file called `compression.cairo`, and we have tried to keep our file structure similar to theirs. So, my suggestions:

* Rename `dict.rs` -> `compression.rs`
* Drop `constants.rs` and include its contents in `compression.rs`

notlesh (approved):

PR #456, opened 2024-12-30
Add support to Pathfinder v0_8

GMKrieger:
Issue Number: N/A

## Type

- [x] feature
- [ ] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Description

Adds support for pathfinder v0_8. It coalesces the 2 types of request on the client and then converts the new version to the one supported by snos.

## Breaking changes?

- [ ] yes
- [x] no


crates/rpc-client/src/client.rs
```diff
@@ -6,6 +6,34 @@ use starknet::providers::JsonRpcClient;
...
+    ReqwestError(#[from] reqwest::Error),
+    #[error("Encountered a version error: {0}")]
+    VersionError(String),
+    #[error("Encountered a custom error: {0}")]
+    CustomError(String),
+}
+
+#[derive(Debug)]
+pub enum ClientVersion {
+    Rpcv07(String),
```
whichqua, line 21: Why do we need the `String` here? The enum should handle it

crates/rpc-client/src/client.rs
```diff
@@ -14,14 +42,19 @@ struct RpcClientInner {
 }
 
 impl RpcClientInner {
-    fn new(base_url: &str) -> Self {
-        let starknet_rpc_url = format!("{}/rpc/v0_7", base_url);
+    fn new(base_url: &str, version: ClientVersion) -> Self {
+        let rpc_version = match &version {
```
whichqua, line 46: Instead of doing this, implement `Display` for `ClientVersion`

crates/rpc-client/src/client.rs
```diff
@@ -6,6 +6,34 @@ use starknet::providers::JsonRpcClient;
...
+    CustomError(String),
+}
+
+#[derive(Debug)]
+pub enum ClientVersion {
+    Rpcv07(String),
+    Rpcv08(String),
+}
+
+impl TryFrom<&str> for ClientVersion {
```
whichqua, line 25: Consider replacing with `impl FromStr` because it allows more flexibility dealing with different string types.

crates/bin/prove_block/tests/prove_block.rs
```diff
@@ -20,75 +20,81 @@ const DEFAULT_COMPILED_OS: &[u8] = include_bytes!("../../../../build/os_latest.j
...
-#[case::key_not_in_proof_2(155172)]
-#[case::l1_gas_and_l1_gas_price_are_0(161476)]
-#[case::key_not_in_proof_3(156855)]
-#[case::key_not_in_proof_4(174968)]
-#[case::timestamp_rounding_1(162389)]
-#[case::timestamp_rounding_2(167815)]
-#[case::missing_constant_max_high(164684)]
-#[case::retdata_not_a_relocatable(160033)]
-#[case::get_tx_info_using_ptr_over_relocatable(243766)]
+#[case::small_block_with_only_invoke_txs(76793, "v0_7")]
```
whichqua, line 23: Does this mean these tests cannot be run on `v0.8`?

crates/rpc-client/src/pathfinder/proofs.rs
```diff
@@ -156,3 +156,112 @@ pub fn verify_proof<H: HashFunctionType>(
...
+        state_commitment: Some(storage_proof.global_roots.block_hash),
+        class_commitment: Some(storage_proof.global_roots.classes_tree_root),
+        contract_proof,
+        contract_data,
+    }
+}
+
+pub fn convert_storage_to_pathfinder_class_proof(storage_proof: StorageProof) -> PathfinderClassProof {
+    let mut class_proof: Vec<TrieNode> = Vec::new();
+    for node in &storage_proof.classes_proof {
```
whichqua, line 253: Consider a more declarative approach
```suggestion
        let class_proof = storage_proof
        .classes_proof
        .iter()
        .map(convert_to_trie_node)
        .collect();
```

crates/rpc-client/src/pathfinder/proofs.rs
```diff
@@ -156,3 +156,112 @@ pub fn verify_proof<H: HashFunctionType>(
...
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize)]
+pub struct StorageProof {
+    pub classes_proof: Vec<NodeHashToNodeMappingItem>,
+    pub contracts_proof: ContractsProof,
+    pub contracts_storage_proofs: Vec<Vec<NodeHashToNodeMappingItem>>,
+    pub global_roots: GlobalRoots,
+}
+
+pub fn convert_storage_to_pathfinder_proof(storage_proof: StorageProof) -> PathfinderProof {
+    let mut contract_proof: Vec<TrieNode> = Vec::new();
```
whichqua, line 207: Consider a more declarative approach
```suggestion
   let contract_proof: Vec<TrieNode> = storage_proof.contracts_proof.nodes
    .iter()
    .map(|node| convert_to_trie_node(node))
    .collect();

let storage_proofs: Vec<Vec<TrieNode>> = storage_proof.contracts_storage_proofs
    .iter()
    .map(|nodes| {
        nodes
            .iter()
            .map(|node| convert_to_trie_node(node))
            .collect()
    })
    .collect();

```

crates/rpc-client/src/pathfinder/proofs.rs
```diff
@@ -156,3 +156,112 @@ pub fn verify_proof<H: HashFunctionType>(
...
+    let mut storage_proofs: Vec<Vec<TrieNode>> = Vec::new();
+    for nodes in &storage_proof.contracts_storage_proofs {
+        let mut trie_nodes: Vec<TrieNode> = Vec::new();
+        for node in nodes {
+            trie_nodes.push(convert_to_trie_node(node));
+        }
+        storage_proofs.push(trie_nodes);
+    }
+
+    let contract_data: Option<ContractData>;
```
whichqua, line 221: This should help you avoid the `if`:
```rs
let contract_data = storage_proof
    .contracts_storage_proofs
    // Get the first "proof" only if it exists
    .get(0)
    // Filter out if the first proof is empty
    .filter(|proof_nodes| !proof_nodes.is_empty())
    .map(|proof_nodes| {
        let first_node = &proof_nodes[0];
        let root = first_node.node_hash;

        match &first_node.node {
            MerkleNode::Binary { left, .. } if left.to_hex_string() == "0x0" => {
                // If left is "0x0", then storage_proofs is empty
                ContractData {
                    root,
                    storage_proofs: vec![],
                }
            }
            // For all other cases (including `MerkleNode::Edge`),
            // store the computed `storage_proofs`.
            _ => ContractData { root, storage_proofs },
        }
    });
```

whichqua (changes requested):
Left some thoughts, let me know what you think.

crates/rpc-client/src/client.rs
whichqua: This is related to https://github.com/keep-starknet-strange/snos/pull/456#discussion_r1900622544

glihm (commented):
@GMKrieger for this PR we may also take in account that we only have one proof now for all the keys, instead of individual proofs, which should change the verification function according to that to accept the union path returned by the RPC call.

crates/bin/prove_block/tests/prove_block.rs
GMKrieger: They should be able to run on v0_8, but they were not made for that, therefore I left them using the old version to maintain their original purpose and added new tests (that repeat some of the old tests that had errors when first developing) on the new api.

ftheirs: I think that it would be good to have a default variable that is set to "v0_7" and maybe put all the "v0_8" tests in a different file / different test. So, once the v0.8 is working for all the tests, we can just replace the default. Besides, it's a little bit cleaner

ftheirs (changes requested):

crates/rpc-client/src/pathfinder/proofs.rs
```diff
@@ -156,3 +156,106 @@ pub fn verify_proof<H: HashFunctionType>(
...
+
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize)]
+pub struct StorageProof {
+    pub classes_proof: Vec<NodeHashToNodeMappingItem>,
+    pub contracts_proof: ContractsProof,
+    pub contracts_storage_proofs: Vec<Vec<NodeHashToNodeMappingItem>>,
+    pub global_roots: GlobalRoots,
+}
+
+pub fn convert_storage_to_pathfinder_proof(storage_proof: StorageProof) -> PathfinderProof {
```
notlesh, line 206: I think this is the wrong idea. The v0.8 API is an improvement over the Pathfinder one, and is more generic, so I think we want to deprecate all of the pathfinder-specific stuff.

notlesh:
Closing this for now. We probably want parts of it but I think we will want to go in the direction of removing all the Pathfinder specific code.

There are also some things we've learned since doing this, namely that proofs are mixed together and are not presented in order like they were in Pathfinder.

PR #460, opened 2025-01-08
Calculate class_commitment from proof

notlesh:
This works around an issue introduced when Pathfinder [removed `class_commitment` from its class proof](https://github.com/eqlabs/pathfinder/pull/2452).

The proofs we get back from `Pathfinder` (prior to v0.8 RPC) are in order. What that means is that the proof is a list of MPT nodes, ordered such that you can walk through the proof from top-to-bottom. This implies that the first node is always the root node, which is the same thing as the `commitment`. In other words, we can get the root/commitment by simply looking at the first node, and that's what this PR does.

Note, however, that this is redundant -- `verify_proof()` ultimately hashes the root node and compares that to `class_commitment`, which is exactly what this PR does. It would be just as good to make `class_commitment` an `Option` an only verify it then; but this PR at least makes the problem clear.

## Type

- [ ] feature
- [x] bugfix
- [ ] build
- [ ] docs
- [ ] testing

## Breaking changes?

- [ ] yes
- [ ] no

### DIFF
--- a/crates/rpc-client/src/pathfinder/proofs.rs
+++ b/crates/rpc-client/src/pathfinder/proofs.rs
@@ -106,6 +106,9 @@ impl PathfinderClassProof {
     ///
     /// However, the proof should always start with the root node, which means all we have
     /// to do is hash the first node in the proof to get the same thing.
+    ///
+    /// NOTE: the v0.8 RPC spec does NOT require the proof to be in order, in which case it is
+    ///       much trickier to guess what the root node is.
     pub fn class_commitment(&self) -> Result<Felt, ProofVerificationError> {
         if !self.class_proof.is_empty() {
             Ok(self.class_proof[0].hash::<PoseidonHash>())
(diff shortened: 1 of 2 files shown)


ftheirs (approved):
LGTM

crates/rpc-client/src/pathfinder/proofs.rs
```diff
@@ -106,6 +106,9 @@ impl PathfinderClassProof {
     ///
     /// However, the proof should always start with the root node, which means all we have
     /// to do is hash the first node in the proof to get the same thing.
+    ///
+    /// NOTE: the v0.8 RPC spec does NOT require the proof to be in order, in which case it is
+    ///       much trickier to guess what the root node is.
```
notlesh, line 111: To elaborate here:

The proof is (1) not required to be in order by the spec and (2) is not in order in my experimentation in `v0.8`. Nor should it be since it consolidates proofs.

A workaround would be to reconstruct a tree view of the proof, in which case it should be clear what the root node is, but this is definitely not ideal.

The `v0.8` RPC proof endpoint does provide commitments, but there are caveats.

whichqua (approved):
LGTM!

HermanObst (approved):

Issue #470, opened 2025-02-13
bug: support deprecated v0 declare txns


whichqua:
**Describe the Feature Request**

Currently we don't support v0 declare txns. 
We need to:
- [x] Support `DECLARE` V0 on snos
- [ ] Fetch and re-execute these txns with blockifier

**Related Code**

```
thread 'main' panicked at ~/snos/crates/rpc-replay/src/transactions.rs:368:42:
not implemented: starknet_rs_to_blockifier with DeclareTransaction::V0
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

**Additional Context**

- https://docs.starknet.io/architecture-and-concepts/network-architecture/transactions/#v0_unsupported_hash_calculation
- https://book.starknet.io/es/ch03-01-transactions.html

whichqua:
We can execute txns on snos successfully.
This test passes:
```rs
#[rstest]
#[tokio::test(flavor = "multi_thread", worker_threads = 1)]
async fn declare_v0_cairo0_account(block_context: BlockContext) {
    use starknet_api::core::{ContractAddress, PatriciaKey};
    use starknet_api::{felt, patricia_key};

    let initial_state = StarknetStateBuilder::new(&block_context).build().await;

    let tx_version = TransactionVersion::ZERO;

    let blockifier_class = BlockifierDeprecatedClass::default();
    let class_info = calculate_class_info_for_testing(blockifier_class.into());

    let sender_address = contract_address!("0x1");
    let declare_tx = blockifier::test_utils::declare::declare_tx(
        declare_tx_args! {
            max_fee: Fee(0),
            sender_address,
            version: tx_version,
            // class_hash: ClassHash(Felt252::from_hex_unchecked("0x5c478ee27f2112411f86f207605b2e2c58cdb647bac0df27f660ef2252359c6")),
        },
        class_info,
    );

    let txs = vec![declare_tx].into_iter().map(Into::into).collect();
    let _result = execute_txs_and_run_os(
        crate::common::DEFAULT_COMPILED_OS,
        initial_state.cached_state,
        block_context,
        txs,
        Default::default(),
        Default::default(),
        HashMap::default(),
    )
    .await
    .expect("OS run failed");
}
```

whichqua:
** Problem 1 **
Implementing the expected converters and trying to execute the txns, the first failure occurs in the following code samples:

1. Class hash for `contract_address!("0x1")` is missing. The default class hash is returned

```rs
[...]

PR #471, opened 2025-02-20
feat: add support for declare v0 txns

whichqua:
Currently we don't support v0 declare txns. 
We need to:
- [x] Support `DECLARE` V0 on snos
- [x] Support and re-execute tx with blockifier
- [x] Include an integration test for `snos`
- [ ]  Include an integration test for `prove_block` 

**Related Code**

```
thread 'main' panicked at ~/snos/crates/rpc-replay/src/transactions.rs:368:42:
not implemented: starknet_rs_to_blockifier with DeclareTransaction::V0
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Issue Number: #470 

## Type

- [x] feature
- [ ] bugfix
- [ ] build
- [ ] docs
- [x] testing

## Breaking changes?

- [ ] yes
- [x] no

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -84,5 +84,5 @@ jobs:
         run: |
           mkdir -p build
           cairo-compile cairo-lang/src/starkware/starknet/core/os/os.cairo --output build/os_latest.json --cairo_path cairo-lang/src
-      - run: cargo +nightly install cargo-udeps --locked
+      - run: cargo install cargo-udeps@0.1.54 --locked
       - run: cargo +nightly udeps --all-targets
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -84,5 +84,5 @@ jobs:
         run: |
           mkdir -p build
           cairo-compile cairo-lang/src/starkware/starknet/core/os/os.cairo --output build/os_latest.json --cairo_path cairo-lang/src
-      - run: cargo install cargo-udeps --locked
+      - run: cargo install cargo-udeps@0.1.54 --locked
       - run: cargo +nightly udeps --all-targets
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -84,5 +84,5 @@ jobs:
         run: |
           mkdir -p build
           cairo-compile cairo-lang/src/starkware/starknet/core/os/os.cairo --output build/os_latest.json --cairo_path cairo-lang/src
-      - run: cargo install cargo-udeps@0.1.54 --locked
+      - run: cargo +nightly install cargo-udeps@0.1.54 --locked
       - run: cargo +nightly udeps --all-targets
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -67,7 +67,7 @@ jobs:
       - uses: actions/checkout@v3
         with:
           submodules: "true"
-      - run: rustup install nightly-2024-05-02
+      - run: rustup install nightly-2024-09-05
       - run: rustup show
       - uses: Swatinem/rust-cache@v2
         with:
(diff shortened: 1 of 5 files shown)


crates/rpc-replay/src/transactions.rs
```diff
@@ -148,6 +148,27 @@ async fn create_class_info(
...
+async fn declare_v0_to_blockifier(
+    tx: &DeclareTransactionV0,
+    client: &RpcClient,
+    block_number: u64,
+) -> Result<blockifier::transaction::transaction_execution::Transaction, ToBlockifierError> {
+    let tx_hash = TransactionHash(tx.transaction_hash);
+    let api_tx = starknet_api::transaction::DeclareTransaction::V0(starknet_api::transaction::DeclareTransactionV0V1 {
+        max_fee: starknet_api::transaction::Fee(felt_to_u128(&tx.max_fee)?),
+        signature: starknet_api::transaction::TransactionSignature(tx.signature.clone()),
+        nonce: starknet_api::core::Nonce(Default::default()),
```
HermanObst, line 160: Write a comment saying that DeclareV0 tx do not have Nonce

HermanObst (approved):
LGTM!

.github/workflows/check.yml
```diff
@@ -84,5 +84,5 @@ jobs:
         run: |
           mkdir -p build
           cairo-compile cairo-lang/src/starkware/starknet/core/os/os.cairo --output build/os_latest.json --cairo_path cairo-lang/src
-      - run: cargo install cargo-udeps --locked
+      - run: cargo install cargo-udeps@0.1.54 --locked
```
notlesh, line 87: Why was this needed?

Re: nightly -- I think it's ok if we run some CI jobs with a nightly compiler (if that helps here), but we don't want the project overall (e.g. in `rust-toolchain.toml`) to require nightly because that forces downstream projects to require nightly.

crates/rpc-replay/src/transactions.rs
notlesh: How are replay attacks avoided if there's no nonce?

tests/integration/common/transaction_utils.rs
```diff
@@ -450,10 +447,11 @@ fn to_internal_l1_handler_tx(l1_tx: &L1HandlerTransaction, chain_id: &ChainId) -
 }
 
 /// Convert a DeclareTransactionV1 to a SNOS InternalTransaction
-pub fn to_internal_declare_v1_tx(
+pub fn to_internal_declare_v0_v1_tx(
```
notlesh, line 450: Make sure the comment here reflects that this works for both versions

tests/integration/declare_txn_tests.rs
```diff
@@ -288,3 +289,44 @@ async fn declare_cairo0_with_tx_info(
...
+            max_fee: Fee(0),
+            sender_address,
+            version: tx_version,
+            class_hash: class_hash.into(),
+        },
+        class_info,
+    );
+
+    let txs = vec![declare_tx].into_iter().map(Into::into).collect();
+    let _result = execute_txs_and_run_os(
```
notlesh, line 321: `_result` is a bit misleading here, as it's not a `Result<_, _>` because of the `expect()` at the end. I'd just leave it as `_` or give it a proper name like `_snos_output`

.github/workflows/check.yml
notlesh: Nevermind, I was thinking this somehow removed `+nightly` but that's on the next line, so we're good.

whichqua: Resolved in https://github.com/keep-starknet-strange/snos/pull/471/commits/29fb4ff3f02d6e2d3f8112d212c1b5b942756ec4

tests/integration/common/transaction_utils.rs
whichqua: Resolved in https://github.com/keep-starknet-strange/snos/pull/471/commits/1583c56687f9878c02c4412c93c5718d2555619a

tests/integration/declare_txn_tests.rs
whichqua: Resolved in https://github.com/keep-starknet-strange/snos/pull/471/commits/963d51c940be0a48e03758c1e3dee6633a75b80c

crates/rpc-replay/src/transactions.rs
whichqua: https://github.com/keep-starknet-strange/snos/pull/471/commits/29a07ee654df20ae282b9bf0e5e8da58bf63f85c

GMKrieger (approved):

PR #473, opened 2025-03-05
test: class hash verification check

whichqua:
This PR tries to deploy a pre_v0.10.0 contract with the different available classhashing approaches and proves that the starknet_core one does not work.

## Type

- [ ] feature
- [ ] bugfix
- [ ] build
- [ ] docs
- [x] testing

## Breaking changes?

- [ ] yes
- [x] no

### DIFF
--- a/tests/integration/class_hash_tests.rs
+++ b/tests/integration/class_hash_tests.rs
@@ -1,0 +1,399 @@
+use std::{
+    collections::HashMap,
+    path::{Path, PathBuf},
+    sync::Arc,
+};
+
+use blockifier::{
+    blockifier::block::BlockInfo,
+    bouncer::BouncerConfig,
+    context::{BlockContext, ChainInfo, FeeTokenAddresses},
+    declare_tx_args, deploy_account_tx_args,
+    execution::contract_class::{ClassInfo, ContractClassV0},
+    invoke_tx_args,
+    state::cached_state::CachedState,
+    test_utils::{
+        create_calldata, declare::declare_tx, deploy_account::deploy_account_tx, invoke::invoke_tx, NonceManager,
+    },
+    transaction::{
+        account_transaction::AccountTransaction,
+        objects::TransactionExecutionInfo,
+        test_utils::{account_invoke_tx, calculate_class_info_for_testing},
+        transactions::ExecutableTransaction,
+    },
+    versioned_constants::VersionedConstants,
+};
+
+use starknet_api::deprecated_contract_class::ContractClass as DeprecatedContractClass;
+
+use blockifier::transaction::transaction_execution::Transaction;
+use cairo_vm::Felt252;
+use rstest::{fixture, rstest};
+use starknet_api::{
+    core::{calculate_contract_address, ChainId, ClassHash, ContractAddress},
+    transaction::{Calldata, ContractAddressSalt, TransactionVersion},
+};
+use starknet_os::{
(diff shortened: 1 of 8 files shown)


tests/integration/class_hash_tests.rs
```diff
@@ -0,0 +1,399 @@
...
+    let execution_infos: Vec<_> =
+        init_txs.into_iter().map(|tx| execute_transaction(tx, &mut cached_state, &block_context)).collect();
+    validate_execution_infos(&execution_infos);
+
+    let mut tx_contracts = vec![];
+
+    let mut txs = Vec::new();
+
+    let (_, pre_0_10_0_contract) = load_cairo0_feature_contract("pre_0_10_0_contract");
+    let pathfinder_class_hash = pre_0_10_0_contract.class_hash().unwrap();
```
ftheirs, line 161: I'd add some comment with links to Pathfinder workaround for the missing compiler version or link the related PRs to this fix

ftheirs (approved):
LGTM

notlesh (changes requested):
This needs some further explanation, esp. in the code.

It sounds like a rather intricate problem, but there isn't enough context in the PR to understand it.

It also needs some rationale in the code, as well. My understanding so far is that we're testing the brokenness of some upstream project and showing in the code that one thing works and the other doesn't.

If that's the right idea, then I could argue that this doesn't belong in our codebase in general. But in any case, it needs some documentation in the tests that explain why it exists here and some more context on the underlying issue.

whichqua:
@notlesh I have added some context here https://github.com/keep-starknet-strange/snos/pull/473/commits/36c33e130a0a82c655b4c93227f9b2ed86bb43c6. Let me know if that helps

tests/integration/class_hash_tests.rs
notlesh: @whichqua Please follow up on this, it's still difficult to understand what all of this PR is doing

tests/integration/class_hash_tests.rs
```diff
@@ -0,0 +1,419 @@
...
+/// - deploy the account
+///
+/// #### Deploy our pre 0.9.0 contract
+/// We will reuse the state for these txns.
+/// We will also attempt to use both hashing mechanisms to assert which snos supports:
+///
+/// - declare the old contract
+/// - deploy the old contract
+///
+/// These tests check which is supported and show that the pathfinder method is the correct one.
```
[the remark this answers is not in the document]
whichqua: Added more context here:
https://github.com/keep-starknet-strange/snos/pull/473/commits/48fe8598a97b32eeac98c9d1c9d1ab61e187b36d

tests/integration/class_hash_tests.rs
whichqua: Added more context with https://github.com/keep-starknet-strange/snos/pull/473/commits/48fe8598a97b32eeac98c9d1c9d1ab61e187b36d

tests/integration/class_hash_tests.rs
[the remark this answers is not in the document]
notlesh: Ok, that helps a lot. One last thing: I'd say up-front which one is broken. The way it's written makes it sound like either one is ok and we just need to pick one, but there's a correct and an incorrect one.

Also, is there an upstream issue for the broken one we can point to?

PR #474, opened 2025-03-05
bug(ci): Also install stable toolchain

notlesh:
This PR attempts to fix an issue where the stable toolchain was not available in CI.

It also disables a test that is currently triggering a bug in `pathfinder`, hopefully this is temporary.


HermanObst (approved):
Good job :)

whichqua (approved):
LGTM!

PR #475, opened 2025-03-14
fix: handle delegate_call syscalls correctly

whichqua:
Currently the `delegate_syscall` had been silenced with a noop `storage_write`. 
This PR fixes that and adds a test in the run_os tests to ensure that the call is handled correctly

This PR also adds `code_addresses` to the map of `accessed_storage_keys` so that the os can successfully access the storage for contracts that were only invoked in a delegate_call.

Issue Number: N/A

## Type

- [ ] feature
- [x] bugfix
- [ ] build
- [ ] docs
- [x] testing

## Breaking changes?

- [ ] yes
- [x] no

### DIFF
--- a/crates/bin/prove_block/src/utils.rs
+++ b/crates/bin/prove_block/src/utils.rs
@@ -94,6 +94,12 @@ pub(crate) fn get_all_accessed_keys(
         }
     }
 
+    let code_addresses = extract_code_addresses(tx_execution_infos);
+
+    for address in code_addresses {
+        accessed_keys_by_address.entry(address).or_default();
+    }
+
     accessed_keys_by_address
 }
 
@@ -136,4 +142,32 @@ fn get_accessed_storage_keys(call_info: &CallInfo) -> HashMap<ContractAddress, H
     }
 
     accessed_keys_by_address
+}
+
+fn extract_code_addresses(transaction_info: &[TransactionExecutionInfo]) -> HashSet<ContractAddress> {
+    let mut addresses = HashSet::new();
+
+    for info in transaction_info {
+        if let Some(call_info) = &info.validate_call_info {
+            extract_inner_addresses(call_info, &mut addresses);
+        }
+        if let Some(call_info) = &info.execute_call_info {
+            extract_inner_addresses(call_info, &mut addresses);
+        }
+        if let Some(call_info) = &info.fee_transfer_call_info {
+            extract_inner_addresses(call_info, &mut addresses);
+        }
+    }
+
+    addresses
+}
+
+fn extract_inner_addresses(call_info: &CallInfo, addresses: &mut HashSet<ContractAddress>) {
+    if let Some(code_address) = &call_info.call.code_address {
+        addresses.insert(*code_address);
+    }
+
+    for inner_call in &call_info.inner_calls {
+        extract_inner_addresses(inner_call, addresses);
+    }
 }
--- a/crates/bin/prove_block/src/utils.rs
+++ b/crates/bin/prove_block/src/utils.rs
@@ -162,6 +162,9 @@ fn extract_code_addresses(transaction_info: &[TransactionExecutionInfo]) -> Hash
     addresses
 }
 
+/// Extracts inner code addresses recursively
+///  
+/// *Note* This is an unbounded recursive call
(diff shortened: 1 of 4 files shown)


crates/starknet-os/src/hints/syscalls.rs
```diff
@@ -61,7 +61,7 @@ where
     let syscall_handler = exec_scopes.get::<DeprecatedOsSyscallHandlerWrapper<PCS>>(vars::scopes::SYSCALL_HANDLER)?;
     let syscall_ptr = get_ptr_from_var_name(vars::ids::SYSCALL_PTR, vm, ids_data, ap_tracking)?;
 
-    syscall_handler.storage_write(syscall_ptr).await?;
```
notlesh, line 64: Weird, I wonder why this was ever doing `storage_write()`...

crates/bin/prove_block/src/utils.rs
```diff
@@ -137,3 +143,31 @@ fn get_accessed_storage_keys(call_info: &CallInfo) -> HashMap<ContractAddress, H
 
     accessed_keys_by_address
 }
+
+fn extract_code_addresses(transaction_info: &[TransactionExecutionInfo]) -> HashSet<ContractAddress> {
```
notlesh, line 147: what is `"code"` in this context?

notlesh: Seems to be the same as contract address, which seems weird to me:

https://docs.rs/blockifier/0.8.0/blockifier/execution/entry_point/struct.CallEntryPoint.html

notlesh: A comment about this might help, otherwise `code_address` appears to be something special...

whichqua: There is `code_address` and `caller_address`. I noticed that the code address is None unless you are doing a delegate_call. This is because you may be calling another contract and hence the caller address would be different.

crates/starknet-os/src/hints/syscalls.rs
whichqua: I actually noticed that I am the one who wrote that, but I was pretty new and I think it was Olivier's idea. You guys were thinking about the replay system right?

ftheirs (approved):

crates/starknet-os/src/execution/deprecated_syscall_handler.rs
```diff
@@ -91,7 +91,7 @@ where
     pub async fn call_contract(&self, syscall_ptr: Relocatable, vm: &mut VirtualMachine) -> Result<(), HintError> {
         self.call_contract_and_write_response(syscall_ptr, CallContract::response_offset(), vm).await
     }
-    #[allow(unused)]
```
notlesh, line 94: This warning that we silenced here might have tipped us off a long time ago that there was a problem :)

crates/bin/prove_block/src/utils.rs
```diff
@@ -137,3 +143,31 @@ fn get_accessed_storage_keys(call_info: &CallInfo) -> HashMap<ContractAddress, H
...
+        }
+        if let Some(call_info) = &info.fee_transfer_call_info {
+            extract_inner_addresses(call_info, &mut addresses);
+        }
+    }
+
+    addresses
+}
+
+fn extract_inner_addresses(call_info: &CallInfo, addresses: &mut HashSet<ContractAddress>) {
```
notlesh, line 165: This is an unbounded (or at least unchecked) recursive call, it would be good to mention that as documentation on the fn. I don't think it's a big deal unless `prove_block` was mission-critical, though.

whichqua, line 165:
```suggestion
/// Extracts inner code addresses recursively 
/// *Note* This is an unbounded recursive call
fn extract_inner_addresses(call_info: &CallInfo, addresses: &mut HashSet<ContractAddress>) {
```

whichqua: Resolved by https://github.com/keep-starknet-strange/snos/pull/475/commits/3ad7e3650da7c78b830e79e8ff34768b3451e2cb

whichqua: Resolved by https://github.com/keep-starknet-strange/snos/pull/475/commits/3ad7e3650da7c78b830e79e8ff34768b3451e2cb

PR #503, opened 2026-02-24
feat(snos): migrate to Starknet 0.14.2 with semantic parity

Mohiiit:
## Summary
This draft migrates SNOS to Starknet `0.14.2` dependency/API surface with semantic parity for OS input assembly.

## Scope
- Upgrades sequencer-related dependencies and Cairo-compatible pins to the 0.14.2 line.
- Adapts transaction/block processing to 0.14.2 type and constructor requirements.
- Removes legacy cached-state input generation path.
- Populates OS block input using real `initial_reads` and computed `block_hash_commitments`.
- Aligns debug serialization to the 0.14.2 input model.
- Adds table-driven coverage (`rstest`) around state-diff conversion and hashing-data assembly.

## Sequence Diagram
```mermaid
sequenceDiagram
    participant RPC as RPC State Source
    participant Exec as Tx Executor
    participant Build as Input Builder
    participant OS as Starknet OS

    RPC->>Exec: Load block context + txs
    Exec->>Exec: Execute txs in canonical order
    Exec->>Build: Execution outputs
    Exec->>Build: Initial reads snapshot
    RPC->>Build: State update
    Build->>Build: Convert to thin state diff
    Build->>Build: Compute block hash commitments
    Build->>OS: Assemble OsBlockInput (0.14.2)
```

## Flow Diagram
```mermaid
flowchart TD
    A[Block + Transactions] --> B[Execute Transactions]
    B --> C[Capture Initial Reads]
    B --> D[Collect Outputs For Hashing]
    A --> E[Fetch State Update]
    E --> F[Convert To ThinStateDiff]
    D --> G[Compute Block Header Commitments]
    F --> G
    C --> H[Assemble OsBlockInput]
    G --> H
    H --> I[Serialize Debug Output]
```

## Notes
- Cutover is 0.14.2-only in this branch (no dual-version shims).
- Migration keeps block-assembly semantics explicit for edge cases (genesis/empty-block/migration vectors).

## Validation Status
- Build, tests, formatting, and lint checks are green in local validation.


crates/core/generate-pie/src/types/block.rs
```diff
@@ -151,39 +158,18 @@ impl BlockData {
...
-            let transaction =
-                transaction.clone().try_into_blockifier_async(&conversion_ctx, trace).await.map_err(|e| {
-                    BlockProcessingError::new_custom(format!(
-                        "Failed to convert transaction to blockifier format: {:?}",
-                        e
-                    ))
-                })?;
+        for (i, transaction) in self.current_block.transactions.iter().enumerate() {
+            let transaction = transaction.clone().try_into_blockifier_async(&conversion_ctx).await.map_err(|e| {
+                BlockProcessingError::new_custom(format!("Failed to convert transaction to blockifier format: {:?}", e))
```
Mohiiit: let's use the proper error handling instead of the new_custom!

crates/core/generate-pie/src/types/block.rs
```diff
@@ -447,7 +511,7 @@ fn populate_alias_contract_keys(
             let contract_felt: Felt = (*(contract_addr)).into();
 
             // Skip if it's address 0x1
-            if contract_felt == Felt::ONE || contract_felt == Felt::TWO {
+            if is_special_contract_felt(contract_felt) {
```
prkpndy: we can add this check before the second for loop

crates/core/generate-pie/src/utils/revert_reason.rs
```diff
@@ -0,0 +1,467 @@
...
+        _ => revert_error.to_string(),
+    })
+}
+
+fn should_strip_vm_tracebacks(error_stack: &ErrorStack) -> bool {
+    has_constructor_frame(error_stack)
+        && error_stack.stack.iter().any(|segment| match segment {
+            ErrorStackSegment::Cairo1RevertSummary(_) => true,
+            ErrorStackSegment::StringFrame(frame) => frame.starts_with("Execution failed. Failure reason:\n"),
+            _ => false,
```
prkpndy, line 39: The receipt commitment hash is sensitive to the exact revert-reason string (the comment on line 9-11 says exactly this). Yet the stripping decision relies on `starts_with` matching a string owned by the Blockifier crate. If Blockifier changes this string even slightly (capitalization, punctuation, trailing newline) in a future upgrade, the normalization will silently stop working and produce receipt hash mismatches — a correctness failure that only surfaces at runtime against real blocks.

The existing tests validate structural properties ("doesn't contain X", "does contain Y") but none assert that the formatted output matches a known golden hash. Consider adding at least one test case that computes the Poseidon hash of the formatted revert reason and asserts it equals a known-good value derived from an on-chain receipt. That way a Blockifier upgrade that changes the string format will fail the test suite immediately rather than silently.

crates/core/generate-pie/src/types/block.rs
```diff
@@ -232,12 +249,26 @@ impl BlockData {
...
             execution_outputs.into_iter().map(|(execution_info, _)| execution_info).collect();
 
+        let mut initial_reads = {
+            let block_state =
+                txn_executor.block_state.as_ref().ok_or(BlockProcessingError::MissingBlockStateAfterExecution)?;
+            capture_extended_initial_reads(block_state)?
+        };
+
         let central_txn_execution_infos: Vec<CentralTransactionExecutionInfo> =
             txn_execution_infos.clone().into_iter().map(|execution_info| execution_info.clone().into()).collect();
```
prkpndy: Redundant `.clone()` — `into_iter()` already consumes the cloned Vec, so each element is owned. The inner `execution_info.clone()` allocates a second copy for nothing:

```rust
// before (two clones per element):
txn_execution_infos.clone().into_iter().map(|execution_info| execution_info.clone().into()).collect();
// after (one clone per element):
txn_execution_infos.clone().into_iter().map(|execution_info| execution_info.into()).collect();
```

`TransactionExecutionInfo` contains full call traces and can be large, so this has real cost on blocks with many transactions.

crates/core/generate-pie/src/types/initial_reads.rs
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    let fetched_entries = fetch_missing_initial_read_storage_entries(state_reader, &missing_entries).await?;
+
+    for (storage_entry, value) in fetched_entries {
+        initial_reads.storage.insert(storage_entry, value);
+    }
+
+    Ok(())
+}
```
prkpndy, line 58: `capture_extended_initial_reads` and `extend_initial_reads_storage` are the two most critical correctness paths in this new module — they determine which storage cells make it into the OS witness. A bug here silently produces an incomplete witness that only fails deep inside OS execution with an opaque error.

Currently only the pure sync helper `collect_missing_initial_read_storage_entries` has a unit test. The two-phase hydration logic (`hydrate_initial_reads` → `get_extended_initial_reads_snapshot`) and the async backfill path (`extend_initial_reads_storage`) have zero test coverage. Consider adding integration-level tests (even with a mock `AsyncRpcStateReader`) that verify:
1. Hydration adds class-hash and nonce entries for every contract seen in initial reads.
2. The second snapshot includes hydrated entries but clears `declared_contracts`.
3. `extend_initial_reads_storage` only fetches entries absent from `initial_reads.storage` and inserts them.

crates/core/generate-pie/src/utils/block_hash.rs
```diff
@@ -0,0 +1,279 @@
...
+    tx_execution_infos: &[TransactionExecutionInfo],
+    thin_state_diff: &ThinStateDiff,
+    l1_da_mode: CoreL1DataAvailabilityMode,
+    starknet_version: &StarknetVersion,
+) -> Result<BlockHeaderCommitments, BlockProcessingError> {
+    let transaction_hashing_data = build_transaction_hashing_data(transactions, tx_execution_infos)?;
+    let (commitments, _measurements) = calculate_block_commitments(
+        &transaction_hashing_data,
+        thin_state_diff.clone(),
+        convert_l1_da_mode(l1_da_mode),
```
prkpndy, line 171: `thin_state_diff.clone()` here is forced by the upstream `calculate_block_commitments` API taking ownership. But `compute_block_hash_commitments` itself takes `&ThinStateDiff`, which means every caller pays for a clone that _looks_ avoidable from the outside.

Two options worth considering:
1. Change `compute_block_hash_commitments` to take `ThinStateDiff` by value (move semantics), letting the caller decide whether to clone. The current sole caller in `block.rs` still needs the diff later, so it would clone there — but the API becomes honest about ownership.
2. If the upstream ever adds a by-reference variant, switch to it.

Minor in isolation, but `ThinStateDiff` can contain the full block's storage diffs which can be substantial.

crates/core/generate-pie/src/types/block.rs
prkpndy: we can move this outside this for loop

crates/rpc-replay/Dockerfile
```diff
@@ -22,8 +22,9 @@ RUN apt-get update && apt-get install -y \
     ca-certificates \
     && rm -rf /var/lib/apt/lists/*
 
-# Install Rust for linux/amd64
-RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain 1.87 \
+# Install Rust for linux/amd64.
+# Keep the image builder aligned with the workspace toolchain used for local validation.
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain 1.89 \
```
prkpndy, line 27: should we update the toolchain version in root Cargo.toml as well?

crates/starknet-os-types/src/casm_contract_class.rs
```diff
@@ -429,7 +432,7 @@ mod tests {
         // Some eccentric type conversions here to load the class hash from string easily, may be
         // improved with more methods on `Hash` / `GenericClassHash`.
         let expected_class_hash =
-            Felt::from_str("0x607c67298d45092cca5b2ae6804373dd8a2cbe7d2ec4072b3f67097461d5ff4").unwrap();
+            Felt::from_str("0x73a7597a1db28cbe2870c4f98d302d4989ad0a12a8ffc441ce965f86a884efd").unwrap();
```
prkpndy: can make this a constant above?

crates/starknet-os-types/src/casm_contract_class.rs
```diff
@@ -517,11 +520,11 @@ mod tests {
     #[rstest]
     #[case::without_segmentation(
         TEST_CONTRACT_WITHOUT_SEGMENTATION,
-        "0xB268995DD0EE80DEBFB8718852750B5FD22082D0C729121C48A0487A4D2F64"
+        "0x19d0d3a56f350492db590c3e79770e0f4edc158a7ea800c2d2e2bf0e57aa08b"
     )]
     #[case::with_segmentation(
         TEST_CONTRACT_WITH_SEGMENTATION,
-        "0x5517AD8471C9AA4D1ADD31837240DEAD9DC6653854169E489A813DB4376BE9C"
+        "0x8dbcbb99355868e8a6147ebe4a29fef7c3c06fa24af109f42d67326799cb07"
```
prkpndy: same

prkpndy (approved):
lgtm. gg 🚀

PR #505, opened 2026-03-25
feat(generate-pie): expose kzg/full-output cli options

Mohiiit:
## Summary
- add `--use-kzg-da` and `--full-output` CLI/env options to `snos`
- wire the CLI values into `OsHintsConfiguration` instead of always using the default helper
- add a small `format_optional_felt` utility helper in `misc`

## Testing
- cargo check -p generate-pie


prkpndy (approved):
lgtm

PR #506, opened 2026-03-30
Support cairo_native in rpc-client

heemankv:
## Summary
- wire `rpc-client`'s existing `cairo_native` feature through to `blockifier/cairo_native`
- handle `RunnableCompiledClass::V1Native` when computing compiled class hash v2
- enable `rpc-client`'s `cairo_native` feature from `generate-pie`
- align SNOS's direct `reqwest` dependency with `starknet-rust-providers` so `rpc-client` can construct a custom `HttpTransport` client
- bound SNOS RPC transport calls with connect/request timeouts so the retry wrapper can recover instead of hanging forever

## Why
Madara now builds `madara` and `orchestrator` in a unified workspace dependency graph. In that graph, `blockifier` is built with `cairo_native`, which exposes `RunnableCompiledClass::V1Native`.

Upstream `rpc-client` still assumed only `V0` / `V1` in `get_compiled_class_hash_v2`, so downstream consumers had to patch SNOS locally. This PR started as the smallest upstream fix needed to let downstreams consume SNOS without a vendored `rpc-client` override.

After wiring Madara directly to this branch and running `orchestrator::tests::jobs::snos_job::test_process_job` against Sepolia block `2934726`, there was still a separate runtime issue: the native path could hang indefinitely on the first `get_nonce_at` state read. The underlying JSON-RPC call returned immediately with `curl`, but SNOS's Rust transport had no bounded connect/request timeout, so a stuck HTTP future never errored and the retry logic in `state_reader` never regained control.

The additional timeout fix in this branch addresses that runtime failure mode. Aligning SNOS's direct `reqwest` dependency with `starknet-rust-providers` was required to use `HttpTransport::new_with_client(...)` with a bounded client.

## Validation
- `cargo check -p rpc-client`
- `cargo check -p rpc-client --features cairo_native`
- `cargo check -p generate-pie`
  - currently blocked in this environment because `apollo_starknet_os_program` requires `cairo-compile` from the sequencer venv/toolchain (`Failed to get cairo-compile version: No such file or directory`)
- Madara exact orchestrator test with SNOS pinned to this branch/commit:
  - `cargo test -p orchestrator --features testing tests::jobs::snos_job::test_process_job -- --exact --nocapture`
  - passed locally once SNOS was consumed from this branch with the timeout change

### DIFF
--- a/crates/rpc-client/src/client.rs
+++ b/crates/rpc-client/src/client.rs
@@ -10,9 +10,13 @@ use starknet_core::types::{ConfirmedBlockId, ContractStorageKeys, StorageKey};
 use starknet_types_core::felt::Felt;
 use std::collections::VecDeque;
 use std::sync::Arc;
+use std::time::Duration;
 
 use crate::constants::{MAX_CONCURRENT_PROOF_REQUESTS, MAX_STORAGE_KEYS_PER_REQUEST, STARKNET_RPC_VERSION};
 use crate::types::{ClassProof, ContractProof};
+
+const RPC_REQUEST_TIMEOUT_SECS: u64 = 15;
+const RPC_CONNECT_TIMEOUT_SECS: u64 = 5;
 
 pub trait ProofClient {
     fn get_proof(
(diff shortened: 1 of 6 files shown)


Mohiiit (approved):

crates/rpc-client/src/client.rs
```diff
@@ -10,10 +10,14 @@ use starknet_core::types::{ConfirmedBlockId, ContractStorageKeys, StorageKey};
 use starknet_types_core::felt::Felt;
 use std::collections::VecDeque;
 use std::sync::Arc;
+use std::time::Duration;
 
 use crate::constants::{MAX_CONCURRENT_PROOF_REQUESTS, MAX_STORAGE_KEYS_PER_REQUEST, STARKNET_RPC_VERSION};
 use crate::types::{ClassProof, ContractProof};
 
+const RPC_REQUEST_TIMEOUT_SECS: u64 = 15;
```
prkpndy, line 18: should we increase it? might be small for larger proofs on slow connections!

prkpndy (approved):
LGTM

heemankv:
Not needed now, already a part of 0.14.2

PR #509, opened 2026-05-04
fix: update Apollo proof facts checks

Mohiiit:
Updates the SNOS-side proof generation stack to the Apollo RC.7 fix release so generated proof facts are aligned with the gateway validation behavior expected for Starknet 0.14.2.

This keeps the proving path and Madara-side validation path on the same Apollo reference.


prkpndy (approved):
lgtm

PR #510, opened 2026-05-12
fix(generate-pie): handle migrated compiled class hashes

Mohiiit:
## What changed
- preserve the pre-migration compiled class hash in initial state reads while still emitting the migrated v2 hash in the OS block input
- validate migrated compiled class hashes reported by RPC against Blockifier before building the block input
- compute CASM `class_hash()` with hash version V1 and add regression coverage for v1/v2 hash divergence

## Why
Migrated classes under SNIP-34 need both hash versions for correct generate-pie execution. The previous flow reused the migrated hash too early, which could feed the wrong compiled class hash into pre-migration reads.

## Impact
Blocks that include migrated compiled classes now keep the correct pre-migration compiled class hash for execution while still serializing migration metadata for downstream consumers.

## Validation
- `CARGO_HOME=/tmp/cargo-home cargo test -p generate-pie -p starknet-os-types` is running in an isolated cargo home because the default `~/.cargo/git` points to a missing external volume on this machine

### DIFF
--- a/crates/core/generate-pie/src/block_processor.rs
+++ b/crates/core/generate-pie/src/block_processor.rs
@@ -119,12 +119,7 @@ pub async fn collect_single_block_info(
     let deprecated_compiled_classes = class_result.deprecated_compiled_classes.clone();
 
     // Step 7b: Extract migrated compiled classes (SNIP-34).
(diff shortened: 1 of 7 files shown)


prkpndy (approved):
LGTM

PR #512, opened 2026-05-17
Make SNOS RPC client and block parallelism configurable on v0.14.2

Mohiiit:
## Summary
- add env-driven SNOS RPC client knobs for request timeout, connect timeout, and `pool_max_idle_per_host`
- add an env-driven cap for generate-pie block parallelism
- stack the change on top of `fix/migrated-compiled-class-input`

## Cairo native note
- `fix/migrated-compiled-class-input` already carries the Cairo native support shape from `codex/rpc-client-cairo-native-minimal`
- this PR adds the runtime env knobs on top of that base

## Env vars
- `SNOS_RPC_REQUEST_TIMEOUT_SECS`
- `SNOS_RPC_CONNECT_TIMEOUT_SECS`
- `SNOS_RPC_POOL_MAX_IDLE_PER_HOST`
- `SNOS_MAX_PARALLEL_BLOCKS`

## Defaults
- request timeout: `60s`
- connect timeout: `5s`
- pool max idle per host: `0`
- max parallel blocks: existing CPU-based default unless overridden

## Validation
- local compile on this branch currently fails in this workstation because the MLIR/LLVM headers for the cairo native toolchain are not fully available (`inttypes.h` missing under the local clang include path)

### DIFF
--- a/crates/rpc-client/src/client.rs
+++ b/crates/rpc-client/src/client.rs
@@ -15,7 +15,7 @@ use std::time::Duration;
 use crate::constants::{MAX_CONCURRENT_PROOF_REQUESTS, MAX_STORAGE_KEYS_PER_REQUEST, STARKNET_RPC_VERSION};
 use crate::types::{ClassProof, ContractProof};
 
-const RPC_REQUEST_TIMEOUT_SECS: u64 = 15;
+const RPC_REQUEST_TIMEOUT_SECS: u64 = 60;
 const RPC_CONNECT_TIMEOUT_SECS: u64 = 5;
 
 pub trait ProofClient {
--- a/crates/rpc-client/src/client.rs
+++ b/crates/rpc-client/src/client.rs
@@ -15,8 +15,12 @@ use std::time::Duration;
 use crate::constants::{MAX_CONCURRENT_PROOF_REQUESTS, MAX_STORAGE_KEYS_PER_REQUEST, STARKNET_RPC_VERSION};
 use crate::types::{ClassProof, ContractProof};
 
-const RPC_REQUEST_TIMEOUT_SECS: u64 = 60;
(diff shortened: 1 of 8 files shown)


crates/rpc-client/src/client.rs
```diff
@@ -76,12 +88,21 @@ impl RpcClientInner {
         let starknet_rpc_url = format!("{}/rpc/{}", base_url, STARKNET_RPC_VERSION);
         info!("Initializing Starknet RPC client with URL: {}", starknet_rpc_url);
 
+        let rpc_request_timeout_secs = Self::read_env_u64(RPC_REQUEST_TIMEOUT_ENV, DEFAULT_RPC_REQUEST_TIMEOUT_SECS);
```
heemankv, line 91: Design: I think this config should stay at the caller boundary rather than be read inside the SNOS library. `generate_pie` is used both from the SNOS CLI and from Madara orchestrator as a library, so reading `SNOS_*` here makes behavior depend on ambient process env and bypasses the existing `clap` / config surfaces on both sides. Could we instead add an explicit runtime config (`RpcClientConfig` / `PieRuntimeConfig` or similar), have the SNOS binaries parse `SNOS_*` with `clap`, and let Madara pass its own orchestrator settings explicitly through the library API? The same concern applies to the new max-parallel-blocks env knob.

Mohiiit: anything works, env sounds easier to me anyway, snos cli is not supposed to be used for prod anyway

heemankv: I checked other codes in orchestrator, there are other many env smells present already, maybe we'll clean up all these together sometime later with orch config feature. good with the changes for now

heemankv (approved):

PR #513, opened 2026-05-18
Add SNOS instrumentation for Patricia mismatches and RPC failures

heemankv:
## Summary
- add contextual logging for cached-state RPC fallbacks that currently collapse to zero values
- log proof fetch, proof repair, and commitment-root summaries around the Patricia input path
- add optional OS input snapshot dumping on failure via `SNOS_DUMP_OS_HINTS_ON_FAILURE`

## Why
We are chasing intermittent SNOS Patricia failures where the OS later trips on a leaf mismatch such as `0 != 1694`. The most suspicious path is silent fallback-to-zero behavior during cached-state reads and other dropped RPC/proof fetches. This PR adds the logs needed to identify the exact contract/key/value path on the next failure.

## Verification
- `cargo fmt --all`
- `PATH="/Users/heemankverma/Work/Carnot/repos/snos/pr512.clone/venv/bin:$PATH" cargo check --workspace`


heemankv:
was needed for debugging, not needed now, closing

PR #514, opened 2026-05-21
Harden proof and class fetch failure handling

heemankv:
## Summary
- retry storage-proof and class-proof fetches before failing
- replace the additional-proof panic with a typed error
- stop silently dropping class fetch failures during state update processing
- include block/address/class context in proof-fetch errors

## Why
`main` already has broader retry coverage and a different pre-state pipeline, so this PR keeps the scope narrow to the remaining proof/class-fetch paths that can still fail opaquely or continue with incomplete data.

## Validation
- cargo fmt --all
- cargo check -p rpc-client

### DIFF
--- a/crates/core/generate-pie/src/state_update.rs
+++ b/crates/core/generate-pie/src/state_update.rs
@@ -446,11 +446,12 @@ async fn process_accessed_addresses(
     info!("Fetched {} contract classes, now compiling in parallel...", class_results.len());
 
     // Phase 3: Compile classes in parallel using rayon (CPU parallelization)
-    let mut successful_class_results = Vec::with_capacity(class_results.len());
-    for (class_hash, contract_class_result) in class_results {
-        let contract_class = contract_class_result.map_err(StateUpdateError::RpcError)?;
-        successful_class_results.push((class_hash, contract_class));
-    }
+    let successful_class_results: Vec<_> = class_results
+        .into_iter()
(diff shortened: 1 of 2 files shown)


crates/core/generate-pie/src/state_update.rs
```diff
@@ -506,13 +507,15 @@ async fn process_accessed_classes(
...
+    }
+
+    let compilation_results: Vec<(Felt252, Result<GenericCompiledClass, StateUpdateError>)> = successful_class_fetches
         .into_par_iter()
-        .filter_map(|(class_hash, contract_class_opt)| {
-            let contract_class = contract_class_opt?;
-            let compiled_result = compile_contract_class(contract_class);
-            Some((class_hash, compiled_result))
-        })
+        .map(|(class_hash, contract_class)| (class_hash, compile_contract_class(contract_class)))
```
prkpndy: any reason to move this to for loop?

prkpndy (approved):
looks good overal. just one question!

crates/core/generate-pie/src/state_update.rs
heemankv: You were right. The `for` loop wasn’t required for the fail-fast behavior; that came from propagating the `Result`, not from the loop itself.
I’ve changed it back to the iterator-style `map(...).collect::<Result<_, _>>()?` form in both places so the code stays cleaner and consistent.

PR #515, opened 2026-05-21
Retry and fail fast on RPC-backed pre-state reads

heemankv:
## Summary
- add a shared RPC retry helper for SNOS Starknet RPC call sites
- stop cached-state construction from converting transport/request failures into zero values
- apply retries to remaining raw block/class/state/proof fetch paths and fail cleanly when retries are exhausted
- stop silently dropping class fetch failures during state update processing and replace the additional-proof panic with a typed error

## Why
Recent SNOS failures showed that transient Pathfinder/RPC failures could be turned into zero-valued pre-state reads on the 0.14.1 branch, which later surfaced as Patricia mismatches. This change makes those failures retryable and then explicit instead of silently corrupting the witness.

## Validation
- cargo fmt --all
- cargo check -p rpc-client

### DIFF
--- a/crates/core/generate-pie/src/cached_state.rs
+++ b/crates/core/generate-pie/src/cached_state.rs
@@ -1,8 +1,8 @@
 use futures::stream::{self, StreamExt};
 use log::info;
 use rayon::iter::{IntoParallelRefIterator, ParallelIterator};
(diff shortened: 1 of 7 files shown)


heemankv:
Not needed now

PR #517, opened 2026-05-25
Keep only storage proof retries from #514

heemankv:
## Summary
- keep only the `get_proof(...)` retry path from #514
- restore pre-#514 behavior for accessed-address and accessed-class fetch handling in `state_update`
- drop the unrelated class-proof retry and state-update fail-fast changes

## Why
Pathfinder storage proof fetches can fail transiently and benefit from retries. The rest of #514 changed class/state-update behavior and surfaced `ClassHashNotFound` failures that were not part of the storage-proof retry fix. This narrows the patch back to the intended proof-retry behavior only.

## Validation
- `cargo fmt --all`
- skipped heavier build validation per request


prkpndy (approved):
lgtm

PR #524, opened 2026-06-15
[codex] fail fast on incomplete RPC data

prkpndy:
## What changed
- stop silently accepting missing or failed `get_class` RPC results while building SNOS inputs
- reject transactions that arrive without `proof_facts` instead of converting them to an empty/default value
- require storage proof fields such as contract roots and class commitment roots instead of fabricating fallback values
- validate alias-contract initial reads against the current block state diff, refetch those reads when they are inconsistent, and fail clearly if they remain inconsistent

## Why
SNOS was still able to continue in a few places after incomplete or inconsistent Pathfinder RPC responses. That let malformed witnesses survive long enough to fail later inside the Cairo OS with Patricia diff asserts, which is much harder to debug.

## Impact
RPC data now has to be present and self-consistent. If Pathfinder does not return a required value, or returns an inconsistent alias-contract pre-state, SNOS will retry where appropriate and then fail early with an explicit error instead of using a default and continuing.

## Validation
- `cargo test -p generate-pie --lib`
- `cargo test -p rpc-client fetch_proof_chunks_retries_only_failed_chunk -- --nocapture`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,8 @@
 
 DEFAULT_NETWORK := sepolia
 DEFAULT_BLOCK_NUMBERS := 924015
+E2E_TEST_THREADS ?= 1
+E2E_SNOS_MAX_PARALLEL_BLOCKS ?= 1
 
 # Starknet Sepolia
 SEPOLIA_RPC_URL ?= https://pathfinder-sepolia.d.karnot.xyz
@@ -143,7 +145,8 @@ test-e2e: ## Run simple parameterized PIE generation tests
 	RUST_LOG=info \
 	SNOS_RPC_URL="$(MAINNET_RPC_URL)" \
 	SNOS_RPC_URL_SEPOLIA="$(SEPOLIA_RPC_URL)" \
-	cargo test -p e2e-tests test_pie_generation -- --nocapture
+	SNOS_MAX_PARALLEL_BLOCKS="$(E2E_SNOS_MAX_PARALLEL_BLOCKS)" \
+	cargo test -p e2e-tests test_pie_generation -- --nocapture --test-threads=$(E2E_TEST_THREADS)
 
 test-all: test-workspace test-e2e ## Run all tests (unit + integration + e2e)
 
--- a/crates/core/generate-pie/src/conversions.rs
+++ b/crates/core/generate-pie/src/conversions.rs
@@ -63,6 +63,8 @@ pub enum ConversionError {
     FieldConversionFailed { field: String, reason: String },
     #[error("Missing transaction receipt for {tx_hash:#x}")]
     MissingTransactionReceipt { tx_hash: Felt },
+    #[error("Missing proof facts in RPC response for transaction {tx_hash:#x}")]
+    MissingProofFacts { tx_hash: Felt },
 }
 
 // ================================================================================================
@@ -306,8 +308,11 @@ pub(crate) fn transaction_receipt_hash(receipt: &TransactionReceipt) -> Felt {
     }
 }
 
-fn proof_facts_from_rpc(proof_facts: Option<Vec<Felt>>) -> starknet_api::transaction::fields::ProofFacts {
-    proof_facts.unwrap_or_default().into()
+fn proof_facts_from_rpc(
+    tx_hash: Felt,
+    proof_facts: Option<Vec<Felt>>,
+) -> Result<starknet_api::transaction::fields::ProofFacts, ConversionError> {
+    proof_facts.ok_or(ConversionError::MissingProofFacts { tx_hash }).map(Into::into)
 }
 
 #[allow(clippy::result_large_err)]
@@ -425,7 +430,7 @@ impl TryIntoBlockifierAsync<TransactionConversionResult> for InvokeTransactionV3
             account_deployment_data: starknet_api::transaction::fields::AccountDeploymentData(
                 self.account_deployment_data,
             ),
-            proof_facts: proof_facts_from_rpc(self.proof_facts),
+            proof_facts: proof_facts_from_rpc(self.transaction_hash, self.proof_facts)?,
         });
 
         let invoke_tx = starknet_api::executable_transaction::InvokeTransaction::create(api_tx, ctx.chain_id)?;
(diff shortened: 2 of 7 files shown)


Makefile
```diff
@@ -3,6 +3,8 @@
 
 DEFAULT_NETWORK := sepolia
 DEFAULT_BLOCK_NUMBERS := 924015
+E2E_TEST_THREADS ?= 1
+E2E_SNOS_MAX_PARALLEL_BLOCKS ?= 1
```
Mohiiit, line 7: Do we need these e2e serialization knobs in this PR? They seem useful for reducing validation flakiness, but they look test-harness specific rather than part of the RPC inconsistency fail-fast change. Would it make sense to either call this out explicitly in the PR description or keep it separate?

crates/core/generate-pie/src/conversions.rs
```diff
@@ -306,8 +308,12 @@ pub(crate) fn transaction_receipt_hash(receipt: &TransactionReceipt) -> Felt {
     }
 }
 
-fn proof_facts_from_rpc(proof_facts: Option<Vec<Felt>>) -> starknet_api::transaction::fields::ProofFacts {
-    proof_facts.unwrap_or_default().into()
+#[allow(clippy::result_large_err)]
```
Mohiiit: Could we avoid adding another bare `#[allow(clippy::result_large_err)]` here? I see this pattern already exists in `generate-pie`, so this may be a broader cleanup, but adding one more unqualified allow makes the lint debt harder to track. Would `#[expect(clippy::result_large_err)]` with a short reason, or a TODO to box/shrink the error type across this module, be a better compromise?

crates/core/generate-pie/src/conversions.rs
```diff
@@ -795,4 +801,36 @@ mod tests {
         assert_eq!(converted_tx.proof_facts_length(), 3);
         assert_eq!(converted_tx.tx_hash().0, tx.transaction_hash);
     }
+
+    #[tokio::test]
```
Mohiiit, line 808: Could we add a small test for `proof_facts: Some(vec![])` succeeding? The RPC spec distinguishes an included empty array from a missing field, so this would document that `None` is invalid after requesting `INCLUDE_PROOF_FACTS`, while an empty array is a valid "no proof facts" response.

crates/core/generate-pie/src/state_update.rs
```diff
@@ -150,6 +150,10 @@ struct CompiledClassBuildInputs<'a> {
     accessed_classes: &'a HashSet<Felt252>,
 }
 
+type AddressBlockLookup = (Felt252, BlockId, bool);
```
Mohiiit: Could we replace the tuple aliases that carry a `bool` with a small struct or enum like `StateScope::{Previous, Current}`? The boolean currently means "is previous", but that intent disappears at call sites, and this path decides which block snapshot we fetch classes from.

crates/core/generate-pie/src/state_update.rs
```diff
@@ -310,6 +314,45 @@ fn extract_migrated_compiled_classes(state_diff: &StateDiff) -> HashMap<Felt252,
...
+    for address in addresses_to_process {
+        if let Some(previous_block_id) = previous_block_id {
+            if let Some(previous_class_hash) = pre_state_class_hashes.get(address) {
+                if previous_class_hash.0 == Felt::ZERO {
+                    debug!(
+                        "Contract {:?} has no class hash in the previous state; skipping previous class fetch",
+                        address
+                    );
+                } else {
+                    class_fetch_pairs.push((*address, previous_block_id, true, previous_class_hash.0));
```
Mohiiit: Can you confirm how this interacts with the later `ContractNotFound` skip for the current-block lookup? I think skipping `ContractNotFound` can be valid for accessed addresses that are touched indirectly but not actually deployed in the current block. The only edge I was wondering about is when we already have a non-zero pre-state class hash for this address, because then current-block `ContractNotFound` would look more like an RPC inconsistency. If that case is impossible in practice, this seems fine; otherwise it might be worth making the distinction explicit.

crates/core/generate-pie/src/types/proof.rs
```diff
@@ -28,6 +28,17 @@ pub struct CommitmentCalculationResult {
     pub contract_class_commitment_info: CommitmentInfo,
 }
 
+fn required_contract_data<'a>(
```
Mohiiit, line 31: Could we make these missing-proof-field errors a bit more structured instead of passing preformatted strings into `new_custom`? Since this PR intentionally fails earlier on malformed RPC proof data, fields like proof side/current-vs-previous, missing field name, and contract address where available would make these failures easier to debug and search for later.

crates/core/generate-pie/src/state_update.rs
```diff
@@ -636,3 +662,46 @@ fn format_declared_classes(state_diff: &StateDiff, class_hash_to_compiled_class_
...
+    #[test]
+    fn collect_fetched_items_propagates_rpc_errors() {
+        let fetched_results = vec![(Felt::ONE, Ok("ok")), (Felt::TWO, Err(ProviderError::RateLimited))];
+
+        let error = collect_fetched_items(fetched_results).expect_err("rpc errors must not be dropped");
+
+        assert!(matches!(error, StateUpdateError::RpcError(ProviderError::RateLimited)));
+    }
+
+    #[test]
```
Mohiiit: Tiny test-structure thought: these two `plan_accessed_address_class_fetches_*` tests look like they could be `rstest` cases over the pre-state class hash and expected fetch plan. Not a blocker, but it might reduce duplicated setup while keeping the behavior names clear.

Mohiiit (commented):
I have a few comments, mostly around keeping the new fail-fast paths easy to debug and avoiding small maintainability debt. The overall direction makes sense to me.

Makefile
prkpndy: I kept this in the same PR because the stricter fail-fast behavior immediately exposed a separate but directly blocking issue in our validation path: `make test-e2e` was driving enough concurrent RPC traffic to the shared Pathfinder that we started getting empty / non-JSON transport responses. Without serializing the e2e harness, the PR would still fail validation, but for RPC saturation noise rather than the fail-fast changes themselves. So this is test-harness specific, but it is coupled to this PR in the sense that it was required to make the new behavior testable and reproducible against the current shared RPC setup.

Mohiiit (approved):
Validated the follow-up changes and local checks pass. Approving.

PR #528, opened 2026-06-25
fix(generate-pie): infer zero-price resource bounds as fee-free

Mohiiit:
## Summary

Teach SNOS replay to classify v3 transactions with zero max-price resource bounds as fee-free.

## Why

Madara bootstrap transactions can carry nonzero resource amounts for execution while using zero prices to represent a fee-free setup transaction. Without this, SNOS asks Blockifier to charge fees and replay fails before Starknet OS execution.

## Validation

The dependent Madara PR passed the sequencer-mode bridge e2e workflow with SNOS fees enabled and this revision pinned.


prkpndy (approved):
LGTM

